PR #2, opened 2014-09-13
Add links to the two explainer docs

AFBarstow:
Added links to the two explainer documents. 

(Not sure why "W3C" needs to be in the title?)

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,8 @@
 editing-explainer
 =================
 
-High level explainer covering the way in which we are addressing editing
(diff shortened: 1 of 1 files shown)


darobin:
I believe "W3C" is in the title so that this can be published as a an extension to HTML, enabling the use of CC-BY.

BenjamP:
Why does W3C need to be in the title for that?

darobin:
Because, sadly, it's a requirement of the CC-BY exemption policy.

That said, feel free to remove it. We can see if it is still needed at the time of publication and add it then!

PR #89, opened 2015-09-25
Firefox 41+ supports the cut and copy commands without setting a pref

cpeterso:
@mystor implemented the `cut` and `copy` commands in Firefox 41 (bug 1012662):

https://bugzilla.mozilla.org/show_bug.cgi?id=1012662


johanneswilm:
Nice! Thanks!

PR #95, opened 2016-01-09
Remove ContentEditableTrue

rniwa:
Remove ContentEditableTrue spec since it doesn't have any content per 1/7 F2F meeting in Redmond.


johanneswilm:
Hey, what F2F meeting was this, and why wasn't this announced on the mailiing list of the taskforce?

I am generally not opposed to removing contentEditable=True, but if I remember correctly last time theme issue was discussed, there were some who felt very strongly that it should not be touched.

johanneswilm:
@rniwa The meeting notes don't seem to say anything about cE=True. Maybe there are some other meeting notes that cover this?

rniwa:
This was a cross-vendor F2F meeting we had in Redmond.  In general, we shouldn't be posting an empty spec with no substance.

johanneswilm:
Hey @rniwa, let's ask the W3C what the proper procedures are.

I disagree that the document was empty and I emailed you and Gary, the organizer of the meeting, explaining the circumstances of the document yesterday. The question of whether contentEditable=True is outdated or not was something that was heavily discussed in the taskforce a few months ago, and there did not seem to be consensus either way.

chaals:
Yeah, the chairs are considering this.

The meeting was not in accordance with W3C process and most particularly does not suggest that any decision has consensus, so this "decision" is under active review.

PR #121, opened 2016-04-11
Insert image in link

caub:


johanneswilm:
As this is just a note about current implementations, I merged it as is.

PR #146, opened 2016-08-24
Change default paragraph separator from <p> to <div>

ayg:
This matches Edge, Blink, WebKit, and possibly soon Gecko
(https://bugzilla.mozilla.org/show_bug.cgi?id=1297414).  At the time the
spec was written, there was no clear-cut behavior, so I went with &lt;p>
because it's more correct semantically, but now we should spec the
behavior that implementations have converged on.

See also w3c/web-platform-tests#3536.

### DIFF
--- a/execCommand.html
+++ b/execCommand.html
@@ -1631,7 +1631,7 @@ data-anolis-spec="dom" href=
         "http://www.whatwg.org/specs/web-apps/current-work/multipage/browsers.html#htmldocument">
         HTMLDocument</a></code> is associated with a string known as the
         <dfn id="default-single-line-container-name">default single-line
-        container name</dfn>, which must initially be "p". (<a href=
+        container name</dfn>, which must initially be "div". (<a href=
         "#the-defaultparagraphseparator-command">The <code title=
         "">defaultParagraphSeparator</code> command</a> can be used to modify
         or query it, by means of the <code><a href=
@@ -16994,6 +16994,10 @@ foo&lt;br&gt;bar
                 discussed in <a href=
                 "http://lists.whatwg.org/htdig.cgi/whatwg-whatwg.org/2011-May/031577.html">
                 a whatwg thread</a>.</p>
+
+                <p>Update August 2016: The consensus of current browsers is now
+                to use div as the separator, not p, so the spec is updated
+                accordingly.</p>
             </div>
 
             <p><a href="#action">Action</a>:</p>


ayg:
+@ayg

ayg:
Marked as non-substantive for IPR from ash-nazg.

ayg:
I got confused at the IPR stuff and I sort-of accidentally said "Mark as non-substantive", which made it happy, so if that's wrong, please fix or tell me what to do.

johanneswilm:
looks good.

Reinmar:
Does choosing `<div>` for a default separator have any effect on other things than `execCommand()`? Or it's strictly related to what `execCommand()` does today?

I'm asking because all editors I know use `<p>`, but also none that I know use `execCommand()`. So if this change is only meant to standardise `execCommand()` I guess we're fine. But if it has any influence on how other things were/will be designed, then it looks bad for me.

johanneswilm:
As I understand it, the change is descriptive rather than normative. In other words: the spec just reflects better what browsers actually do already, not the other way round.

Given that the browser makers have declared that they won't change their browsers due to things in this spec and that they will just try to have their browsers behave more alike, I think it's fine.

I also agree with @Reinmar that in practice no serious editor is currently using execCommand. That may or may not change in the future, but if it is to change, probably the browser makers will have to put a lot of resources into this area. It is not currently a priority.

ayg:
It also affects what happens when the user presses "Enter" (which is equivalent to calling execCommand("insertparagraph")).  Do any editors use this, or do they all override it?  I assume they override it, and would be very interested to know if they don't (because it means the code might be significant).

Either way, as @johanneswilm says, this is just reflecting what browsers have already done.  If you do want to use built-in browser handling for the user hitting Enter, and want it to produce <p>, you can call document.execCommand("defaultparagraphseparator", false, "p"), although it will not work in current Firefox (although hopefully it will in a few versions).

Actually, though, it looks like I was wrong.  Edge does still produce &lt;p>.  I don't know what happened, I was sure I got it producing &lt;div>.  This change may still be good, though, because it's still what Blink/WebKit do, and I was told they're not likely to change, and with default styles &lt;div> is closer to what Firefox now does.  So Edge is still the only one out.  I guess I'll have to figure out what Firefox will do before deciding whether to make a new PR.

johanneswilm:
Yes, do you have ani insights into why they are changing this around right now?

ayg:
"Yes" to which?

I was working on a few Firefox editor bugs and always thought this would be a good one to tackle, because the way Firefox generates line breaks is drastically out of step with other UAs, and if you don't override it, it changes what happens every time the user hits enter.  With this change, you can run execCommand("defaultparagraphseparator", false, "p") and then all browsers will output &lt;p>, which can't hurt if anyone is actually using the default Enter handling.

I would be very interested in knowing what parts of contenteditable are not overridden in common use, by the way.  E.g., where the cursor goes on clicks/key movements, default behavior for backspace/enter/etc., things like that.  If there are things that I know affect sites, it would make sense to prioritize them.

johanneswilm:
Yes to "This change may still be good"

johanneswilm:
Caret movement is not overriden in some editors, while others do. One thing that is super problematic is when the caret refuses to be displayed certain places or cannot go there at all, not even programmatically. That also applied to Firefox last time I checked and is the reason we added a list of "legal caret positions" to the new cE-spec.

ayg:
Do editors all override hitting Return, Backspace, etc.?

If there are any Firefox bugs you think are particularly worthwhile to fix, please file a few and CC me (or give me a link if you already did).  I can't promise that I'll be able to fix them, but there's a chance.  (Unfortunately, I don't know anything about our caret positioning code, so probably I won't be able to handle that.)

johanneswilm:
> Do editors all override hitting Return, Backspace, etc.?

Those involve taking apart and merging paragraphs, so I cannot really imagine how one would create an editor without doing that. At the very least one will need to monitor what the browser does and then adjust if needed. Not sure what @Reinmar think, but I would say for those just try to be as similar as Chrome given that that is the currently most popular browser on those. Given that the Chrome team says they are not going to fix this, the best we can hope for is that all browsers will be broken in the same or a similar way.

> If there are any Firefox bugs you think are particularly worthwhile to fix, please file a few and CC me (or give me a link if you already did).

I linked to two bug reports @Reinmar and me filed in a 2014 ACM article [1]. My top priority bug for Firefox is the one mentioned above, and was filed in 2013 [2]. I can also see that @Reinmar has filed a number of issues in the period 2013-15 [3]. Most of that still seems current, but priorities may have shifted slightly on the part of JS editors due to the changes in how JS editors are operating nowadays. The top priority right now would really seem to be to get the beforeinput event implemented in Firefox in the same way that Chrome is currently implementing it. Or what would you say, @Reinmar?

[1] http://dl.acm.org/citation.cfm?id=2723154
[2] https://bugzilla.mozilla.org/show_bug.cgi?id=873883
[3] https://www.google.com/search?q=Piotrek+Koszuli%C5%84ski&gws_rd=ssl#safe=off&q=piotrek+koszuli%C5%84ski+site:bugzilla.mozilla.org

ayg:
If people can give me a prioritized list of bugs to look at, that would be helpful.  Things that are focus/caret-related will require some work on my part to get into, but it might be a good idea for me to try anyway.  In theory beforeinput doesn't seem too hard, but it doesn't seem like there's a spec that's precise enough to implement interoperably, is there?  If you want it to be like Chrome does it, it would help a lot to link to the bug where Chrome implemented it, so I can look at their code.  (I looked briefly and didn't find it.)

johanneswilm:
Chrome is currently implementing it so we are adjusting the Input Events spec as we go along. Last I talked to them they thought they should be close to shipping it within the next 3 months. We just had a F2F just about it in San Francisco. Firefox wasn't present as the only browser, for some reason.

ayg:
Do you have a link to the bugs where it's being implemented, or at least the names of the people who are doing it so I can get in touch with them?

johanneswilm:
@choniong is quite active in https://github.com/w3c/input-events/issues and he is part of the team implementing this in Chromium.

dtapuska:
@ayg You can look for the references following the createBeforeInput method here:
https://cs.chromium.org/chromium/src/third_party/WebKit/Source/core/events/InputEvent.h?l=90

chong-z:
Also here is the [tracking bug](https://crbug.com/585875) if it helps, and there are some individual bugs blocking it.

Reinmar:
> It also affects what happens when the user presses "Enter" (which is equivalent to calling execCommand("insertparagraph")). Do any editors use this, or do they all override it?

CKEditor overrides it totally. I think that some editors may try to post-fix this (so base on the native "insertparagraph" algorithm, but change the created block afterwards), but those should also be safe.

> Do editors all override hitting Return, Backspace, etc.?

CKEditor 4 overrides entire enter key behaviour and backspace in certain cases (mainly in lists, tables and in Chrome when it comes to [merging blocks](http://dev.ckeditor.com/ticket/9998)). CKEditor 5 overrides everything.

So: 

> Not sure what @Reinmar think, but I would say for those just try to be as similar as Chrome given that that is the currently most popular browser on those

This wouldn't be fully true for us – as I mentioned above, Chrome has certain behaviours which we cannot accept. But it's hard for me to tell which browser's behaviour would suit us the most, cause we override too many so I don't even know how certain browsers behave :).

> The top priority right now would really seem to be to get the beforeinput event implemented in Firefox in the same way that Chrome is currently implementing it. Or what would you say, @Reinmar?

I agree. Beforeinput would be the biggest improvement for us. In CKEditor 5 we intercept all input, because it must be carefully applied in the custom data model (which is based on operational transformation). Right now we intercept the input based on available events (e.g. `paste`), but most of the typing can only be intercepted by mutation observers. This is pretty tricky, though, if you consider composition, autocorrection, spell checkers, etc. and `beforeinput` would a huge relief :).

Actually, once we're done with the current sprint, we hope to create a PoC of https://github.com/ckeditor/ckeditor5-typing working based on `beforeinput` available in Chrome.

johanneswilm:
> This wouldn't be fully true for us – as I mentioned above, Chrome has certain behaviours which we cannot accept. But it's hard for me to tell which browser's behaviour would suit us the most, cause we override too many so I don't even know how certain browsers behave :).

@Reinmar Are you referring to the `<div>` vs `<p>` issue? Is there anything about Chrome's behavior that is problematic for you that you cannot either already override or override once we have the beforeinput event?

johanneswilm:
I know there is for me there is one issue with Chrome that is almost the same issue there is for Firefox: places that the caret refuses to go [1]. But there is nothing I can think of that involved changing the DOM.

[1] https://bugs.chromium.org/p/chromium/issues/detail?id=242110#

Reinmar:
> @Reinmar Are you referring to the `<div>` vs `<p>` issue? Is there anything about Chrome's behavior that is problematic for you that you cannot either already override or override once we have the beforeinput event?

I was referring to http://dev.ckeditor.com/ticket/9998. I see that the behaviour for merging blocks was finally fixed (no inline styles are created). But I can see that e.g. this is still an issue: http://dev.ckeditor.com/ticket/9998#Backspaceunabletoremoveformatting. However, I think that all this can be overridden by us already and with beforeinput it should be even simpler and more stable than now.

plehegar:
Marked as non-substantive for IPR from ash-nazg.

PR #155, opened 2016-11-28
execCommand: Fix HTML spec URLs

### LINKED ISSUE #154: Broken link in execCommand
In execCommand [6. Common definitions](https://w3c.github.io/editing/execCommand.html#common-definitions), there is a link to `designMode` (http://www.whatwg.org/specs/web-apps/current-work/multipage/editing.html#designMode), but that link is broken:

> The page you are looking for is no longer available at this URL.
> 
> Links to the multipage version of the specification are unfortunately likely to break over time. You might be able to find what you want from the contents page.
> 
> If you have found a broken link on the WHATWG site itself, please e-mail Ian Hickson. If you found a broken link from another site pointing to the WHATWG site, please let that site know of the problem instead. Thanks!

Perhaps it should be changed to https://html.spec.whatwg.org/multipage/interaction.html#designMode

rakuco:
The spec was pointing to old URLs and relying on whatwg.org redirecting to
the right location by parsing a JSON with all new page names. Instead, just
update the spec to point to the final, working URLs directly.

Fixes #154.

### DIFF
--- a/execCommand.html
+++ b/execCommand.html
@@ -99,10 +99,10 @@ var respecConfig = {
             response to execCommand() so we need to treat it as
             presentational", so it includes things like <code class="external"
             data-anolis-spec="html" title="the strong element"><a href=
-            "http://www.whatwg.org/specs/web-apps/current-work/multipage/text-level-semantics.html#the-strong-element">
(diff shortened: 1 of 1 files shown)


siusin:
Marked as non-substantive for IPR from ash-nazg.

PR #197, opened 2019-07-28
[contentEditable] fix ReSpec warnings

saschanaz:


contentEditable.html
```diff
@@ -52,6 +52,7 @@
       <p>
       </p>
     </section>
+    <section id='conformance'></section>
```
saschanaz: @marcoscaceres Is this the right thing to do here? We implicitly mark all sections as informative without this and it causes some warnings, maybe ReSpec should give some guide for this.

marcoscaceres: Yes to both... we need to detect if the document is on the W3C Rec track, and if it's not a note. And then including a conformance section should be sufficient. 

However, could you please move the conformance section to the end of the document? Just before any appendix.

marcoscaceres (approved):

johanneswilm:
Are waiting for something to merge this, @marcoscaceres ?

PR #202, opened 2019-08-25
[execCommand] use inline element syntax

saschanaz:
Batch-converted with `<code[\s\n]*class=[\s\n]*"external"[\s\n]*data-anolis-spec=[\s\n]*"html"[\s\n]*title=[\s\n]*"(?:the )?\w+(?: element)?"><a[\s\n]*href=[\s\n]*[^>]+>[\s\n]*(\w+)[\s\n]*</a>[\s\n]*</code>`.

### DIFF
--- a/execCommand.html
+++ b/execCommand.html
@@ -98,13 +98,7 @@ var respecConfig = {
             it to a span, if it's some type of presentational element.
             ("Presentational" here really means "browsers produce it in
             response to execCommand() so we need to treat it as
-            presentational", so it includes things like <code class="external"
-            data-anolis-spec="html" title="the strong element"><a href=
-            "https://html.spec.whatwg.org/multipage/semantics.html#the-strong-element">
-            strong</a></code> and <code class="external" data-anolis-spec=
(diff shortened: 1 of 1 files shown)


saschanaz:
@sidvishnoi, any idea why `[^sub^]` and `[^sup^]` don't work? Strangely https://respec.org/xref shows nothing for them.

sidvishnoi:
They are not in our data base because HTML spec didn't export them.

saschanaz:
Should an issue be filed on HTML? I don't see why the difference happens because I don't see any explicit export on [`<strong>`](https://html.spec.whatwg.org/multipage/text-level-semantics.html#the-strong-element).

marcoscaceres:
Yeah, we should make sure they get exported from HTML. Could you please file a bug?

saschanaz:
Filed https://github.com/whatwg/html/issues/4863. I don't think it's a blocker though, this already is a big change.

marcoscaceres (approved):
Agree. This is just non-normative ReSpec refactor.

PR #204, opened 2019-08-26
[execCommand] use inline IDL reference for DOM

saschanaz:
This doesn't cover any methods because it looks more complicated.

### DIFF
--- a/execCommand.html
+++ b/execCommand.html
@@ -1036,9 +1036,7 @@
     <section>
         <h2 id="common-definitions">Common definitions</h2>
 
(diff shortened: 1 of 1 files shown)


marcoscaceres (approved):

marcoscaceres:
Methods should just be {{Foo/bar()}}

saschanaz:
> Methods should just be {{Foo/bar()}}

xref only allows full signature with arguments, so it didn't work. Also the spec currently does some `Foo.bar(localVariable1, localVariable2)`.

marcoscaceres:
Ah, ok... we will need to work that out.

PR #206, opened 2019-09-01
[execCommand] Replace Editing[Before]InputEvent to InputEvent

### LINKED ISSUE #200: Should execCommand dispatch beforeinput or not?
I briefly tested `execCommand` and `beforeinput` in https://codepen.io/SaschaNaz/pen/WNepGPa. For `execCommand("bold")`, Chrome does not dispatch `beforeinput` while Safari does dispatch it with `inputType: ""` (whereas both dispatch `input`). As I see nothing about `beforeinput` on execCommand spec, which is the expected behavior here?

### LINKED ISSUE #201: How should execCommand map to inputType?
A simple behavior test in https://codepen.io/SaschaNaz/pen/WNepGPa says that both Firefox and Chrome dispatches `input` with `inputType: "formatBold"` while Safari dispatches with `inputType: ""` for the same `execCommand("bold")`. As there is no standard mapping between `execCommand` commands and `inputType` of `InputEvent`, what's the expected behavior here?

saschanaz:
Closes #201

This intentionally skips describing what should be `inputType` of `beforeinput` per https://github.com/w3c/editing/issues/200#issuecomment-584224961

### DIFF
--- a/ActiveDocuments/execCommand.html
+++ b/ActiveDocuments/execCommand.html
@@ -716,13 +716,13 @@
         </li>
         <li>If the previous step returned false, return false.
         </li>
-        <li>If the action modified DOM tree, then [=Dispatch=] an [= event=] at
+        <li>If the action modified DOM tree, then [=dispatch=] an [=event=] at
         <var>affected editing host</var> using {{InputEvent}} interface. The
         event's {{Event/type}} attribute must be initialized to "input"; its
         {{Event/isTrusted}} and {{Event/bubbles}} attributes must be
         initialized to true; its {{InputEvent/inputType}} attribute must be
-        initialized to the <a data-lt="map an edit command to input type value"
-        >mapped value</a> of <var>command</var>; and its {{InputEvent/data}}
+        initialized to the [=map an edit command to input type value|mapped
+        value=] of <var>command</var>; and its {{InputEvent/data}}
         attribute must be initialized to null.
         </li>
         <li>Return true.
--- a/ActiveDocuments/execCommand.html
+++ b/ActiveDocuments/execCommand.html
@@ -607,44 +607,6 @@
       <h2 id="methods-to-query-and-execute-commands">
         Methods to query and execute commands
       </h2>
-      <p class="XXX">
-        We fire events as requested in <a href=
-        "https://www.w3.org/Bugs/Public/show_bug.cgi?id=13118">bug 13118</a>.
-        This is a new feature does not currently match any browser. <strong>If
-        you are implementing this, please make sure to file any feedback as
-        bugs. The spec is not finalized yet and can still be easily
-        changed.</strong>
-      </p>
-      <pre class="idl">
-            [Exposed=Window,
-             Constructor(DOMString type, optional EditingBeforeInputEventInit eventInitDict = {})]
-            interface EditingBeforeInputEvent : Event {
-              readonly attribute DOMString command;
-              readonly attribute DOMString value;
-            };
-            
-            dictionary EditingBeforeInputEventInit : EventInit {
-              DOMString command = "";
-              DOMString value = "";
-            };
-
-            [Exposed=Window,
-             Constructor(DOMString type, optional EditingInputEventInit eventInitDict = {})]
-            interface EditingInputEvent : Event {
-              readonly attribute DOMString command;
-              readonly attribute DOMString value;
-            };
-            
-            dictionary EditingInputEventInit : EventInit {
-              DOMString command = "";
-              DOMString value = "";
-            };
-        </pre>
-      <p class="XXX">
-        We have two different interfaces because we might want to add
-        additional members to the input event but not the beforeinput event,
-        such as a list of nodes that were affected.
-      </p>
       <p class="note">
         TODO: Define behavior for <var title="">show UI</var>.
       </p>
@@ -4784,7 +4746,7 @@
         </h3>
         <div class="note">
           <p>
-            Color interpretations: 
+            Color interpretations:
             <!-- I wanted the table to stretch out to the left.  CSS doesn't support this
 directly, but where there's a will, there's a way! -->
           </p>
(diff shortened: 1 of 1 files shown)


johanneswilm (changes requested):
I don't think this should be removed. Instead it should be adjusted to comply with the current terminology sued in the Input Events spec. The reason is that execCommand is a highly controversial technology that has fallen out of use a decade ago and that is not likely to be used by editors in the future either. So everything execCommand-related should be contained in this spec and not flow out into other documents.

gked:
@saschanaz are you ok if I close this PR?
@johanneswilm as FYI

saschanaz:
I think we should remove the IDL definition unless any browser has an intention to implement.

johanneswilm:
I think the IDL should be removed as this is not what was ultimately used. But the feature should not be removed. It is implemented in Safari (beforeinput event for execCommand) and it sounds like there is interest in also having this behavior in other browsers. See latest comments in https://github.com/w3c/editing/issues/200

saschanaz:
Will try finding some time to modify this.

johanneswilm (approved):
This looks godo to me. Good work!

ActiveDocuments/execCommand.html
```diff
@@ -710,17 +672,12 @@ <h2 id="methods-to-query-and-execute-commands">
...
-            The event's {{Event/type}} attribute must be initialized to
-            "beforeinput"; its {{Event/isTrusted}}, {{Event/bubbles}}, and
-            {{Event/cancelable}} attributes must be initialized to true; its
-            <code title="dom-EditingBeforeInputEvent-command"><a href=
-            "#dom-editingbeforeinputevent-command">command</a></code> attribute
-            must be initialized to <var title="">command</var>; and its
-            <code title="dom-EditingBeforeInputEvent-value"><a href=
-            "#dom-editingbeforeinputevent-value">value</a></code> attribute
-            must be initialized to <var title="">value</var>.
+            <li>[=Dispatch=] an [=event=] at <var>affected editing host</var>
```
marcoscaceres: Nit: I wonder if we should use the "fire an event" algo here instead? Also, does all this happen sync or at these queued tasks (and if so, what task source)?

ActiveDocuments/execCommand.html
```diff
@@ -759,23 +716,49 @@ <h2 id="methods-to-query-and-execute-commands">
         </li>
         <li>If the previous step returned false, return false.
         </li>
-        <li>If <var title="">command</var> is not in the <a href=
-        "#miscellaneous-commands">Miscellaneous commands</a> section, then
-        [=Dispatch=] an [= event=] at <var title="">affected editing host</var>
-        that uses the <a>EditingInputEvent</a> interface. The event's
-        {{Event/type}} attribute must be initialized to "input"; its
+        <li>If the action modified DOM tree, then [=Dispatch=] an [= event=] at
```
marcoscaceres:
```suggestion
        <li>If the action modified DOM tree, then [=dispatch=] an [=event=] at
```

ActiveDocuments/execCommand.html
```diff
@@ -759,23 +716,49 @@ <h2 id="methods-to-query-and-execute-commands">
...
+        event's {{Event/type}} attribute must be initialized to "input"; its
         {{Event/isTrusted}} and {{Event/bubbles}} attributes must be
-        initialized to true; its <code title=
-        "dom-EditingInputEvent-command"><a href=
-        "#dom-editinginputevent-command">command</a></code> attribute must be
-        initialized to <var title="">command</var>; and its <code title=
-        "dom-EditingInputEvent-value"><a href=
-        "#dom-editinginputevent-value">value</a></code> attribute must be
-        initialized to <var title="">value</var>.
+        initialized to true; its {{InputEvent/inputType}} attribute must be
```
marcoscaceres: NIT: DOM spec defines a nice hook algorithm "event construction steps"... you don't need to use them, but they might be useful here if you'd like. Maybe a followup :)

ActiveDocuments/execCommand.html
```diff
@@ -759,23 +716,49 @@ <h2 id="methods-to-query-and-execute-commands">
...
+        initialized to the <a data-lt="map an edit command to input type value"
+        >mapped value</a> of <var>command</var>; and its {{InputEvent/data}}
```
marcoscaceres:
```suggestion
        initialized to the [=map an edit command to input type value|mapped
        value=] of <var>command</var>; and its {{InputEvent/data}}
```

marcoscaceres (approved):
couple of nits and suggestions for better DOM spec integration.

ActiveDocuments/execCommand.html
saschanaz: Looking at [your work](https://w3c.github.io/payment-request/#merchantvalidationevent-constructor), it seems it's defined for each constructor. Since we are reusing an event interface from another spec, is it still applicable here?

marcoscaceres: ah, you are right and great point. as it's just Event, it doesn't need any special set up.

saschanaz:
@johanneswilm Per [Input Event](https://w3c.github.io/input-events/#event-definitions) the default action of `beforeinput` is set to none when it's not insert/deleteCompositionText. Should it be fixed too?

ActiveDocuments/execCommand.html
saschanaz: >Also, does all this happen sync or at these queued tasks (and if so, what task source)?

It happens synchronously on both Firefox and Chrome. Check out https://codepen.io/SaschaNaz/pen/MWaXLQE to see the behavior.

saschanaz:
Thank you all for your reviews!

johanneswilm:
@saschanaz I think that says:

> 'Update the DOM element' for contentEditable="true" editing hosts for all inputTypes.

The line you are referring to is this one I believe:

> 'Update the DOM' for contentEditable=typing editing hosts for inputTypes "insertCompositionText" and "deleteCompositionText".

Given that contentEditable=typing has been removed (for the time being) from all W3C documents, maybe we need to remove it here as well?

PR #229, opened 2020-02-10
editorial: remove editing host definition and update references to it

janiceshiu:
Hi! 
As mentioned in this [HTML spec PR](https://github.com/whatwg/html/pull/5253/files), I am migrating the definition of some terms from the execCommand spec to the HTML spec. 

execCommand's definition of `editing host` has been [merged](https://github.com/whatwg/html/pull/5253/files) into the HTML spec. 

Following @marcoscaceres' [advice](https://github.com/whatwg/html/pull/5253#issuecomment-583948517), I am now removing the `editing host` definition from execCommand and updating execCommand's references to that definition.
We can now use Respec to [cite](https://respec.org/xref/?term=editing+host) `editing host` with `<a data-cite="html">editing host</a>`.

### DIFF
--- a/execCommand.html
+++ b/execCommand.html
@@ -1063,19 +1063,6 @@
         title="concept-node">node</a> that is not a <a href="#block-node">block
         node</a>.</p>
 
-        <p>An <dfn id="editing-host">editing host</dfn> is a <a class=
-        "external" data-anolis-spec="dom" href=
-        "http://dom.spec.whatwg.org/#concept-node" title=
-        "concept-node">node</a> that is either an <a href="#html-element">HTML
-        element</a> with a <code class="external" data-anolis-spec="html"
-        title="attr-contenteditable"><a href=
-        "http://www.whatwg.org/specs/web-apps/current-work/multipage/editing.html#attr-contenteditable">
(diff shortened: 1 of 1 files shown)


execCommand.html
```diff
@@ -602,10 +602,10 @@ <h3 id="enabled-commands">Enabled commands</h3>
             "command">commands</a> defined here are <a href=
             "#enabled">enabled</a> if the <a href="#active-range">active
             range</a> is not null, its [=range/start node=] is either <a href=
-            "#editable">editable</a> or an <a href="#editing-host">editing
+            "#editable">editable</a> or an <a data-cite="html">editing
```
marcoscaceres: You can move the `data-cite="html"` up to the `<body>`, then you only need to mention it once.

marcoscaceres: Also, editing is just a concept, so actually you can do: [=editing host=] and [=editing hosts=] and get rid of the `<a>` everywhere.

marcoscaceres:
Marked as non substantive for IPR from ash-nazg.

marcoscaceres (approved):
Aaaaawsome!

PR #233, opened 2020-03-04
Restructuring the repo for easier navigation

### LINKED ISSUE #232: Restructure the repo a bit for easy navigation

gked:
Closes #232 

Implementation commitment:


johanneswilm (approved):
In general this looks good to me. Note that several of these files really only contain a link to another file. These link files were created whenever a document changed name, so that links linking to that file would not end up being dead. This is a practice I took over from those who maintained this repository before me.

If W3C rules and common practices do not require us to keep old links intact, then we might just as well delete all those files that just contain links. For example, contentEditableTyping/Events/Intentions/etc. were all the same document that just changed its name over time as our focus shifted.

gked:
> In general this looks good to me. Note that several of these files really only contain a link to another file. These link files were created whenever a document changed name, so that links linking to that file would not end up being dead. This is a practice I took over from those who maintained this repository before me.
> 
> If W3C rules and common practices do not require us to keep old links intact, then we might just as well delete all those files that just contain links. For example, contentEditableTyping/Events/Intentions/etc. were all the same document that just changed its name over time as our focus shifted.

sounds good. We can check with Web Apps WG Chairs whether we should delete them or not.

annevk:
This broke a bunch of links, please revert this.

johanneswilm:
@annevk Is it ok if I readd all those link files so that clicking on any old link will automatically transport you to the new location of the file?

PR #242, opened 2020-04-16
execCommand: fix link to contentEditable

gnprice:
This relative link worked before the repo reorg in bd52f3cf2.

To fix, it'd be enough to prepend "../Graduated/" to the link. But
while here we might as well dereference that page's redirect, too.


gnprice:
(I think this edit qualifies as non-substantive.)

saschanaz (approved):

PR #244, opened 2020-05-03
Undeprecate execCommand("useCSS")

saschanaz:
Since useCSS is supported everywhere and we have no plan to remove it, user agents should not emit a deprecation warning.

 * [x] Confirmed there are no *new* ReSpec errors or warnings.

Implementation commitment:

 * [x] WebKit (https://bugs.webkit.org/show_bug.cgi?id=36683)


johanneswilm:
@saschanaz Should we put this on the agenda for the call next week so that we can get a decision on whether or not to move ahead with this?

saschanaz:
Sounds good to me, @masayuki-nakano does this patch sound good to you?

johanneswilm:
@saschanaz Are you saying that we should readd `useCSS` because there is an advantage of using it over `styleWithCSS`? Or do you want it to say that `styleWithCSS` is still the preferred way and `useCSS` should just be listed as non-deprecated because it actually is still working in all current browsers (as a reverse alias for `styleWithCSS`)?

saschanaz:
The latter. I just want to remove the deprecation notice because AFAIK that's generally for features that will disappear anytime soon, which is not the case here. (Unless anyone wants to remove it)

johanneswilm:
I would personally be in favor of adding a small note pointing out that it is doing the same as `styleWithCSS` but with reverse values, and otherwise to "undeprecate" it the way you have done here - unless it is doing something else in some situations.

travisleithead:
WG Call: Ask @saschanaz to files bugs on implementations that may be firing deprecation warnings, and then we'll merge the PR.

saschanaz:
> 
> 
> WG Call: Ask @saschanaz to files bugs on implementations that may be firing deprecation warnings, and then we'll merge the PR.

Confirmed that Firefox, Chrome, and WebKit (on Windows build) all fire no warning for `document.execCommand("useCSS", false, false)`.

johanneswilm (approved):

PR #245, opened 2020-05-03
Fix execCommand link

### LINKED ISSUE #247: execCommand archive link is broken
https://w3c.github.io/editing/#archive-3 includes a link to execCommand that doesn't work.

saschanaz:
Since it's in ActiveDocuments directory.

Fixes #247

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,22 +1,22 @@
 Editing Task Force
 =================
 
-Editing Task Force is a part of [W3C WebApps Working Group](https://w3c.github.io/webappswg/). 
+Editing Task Force is a part of [W3C WebApps Working Group](https://w3c.github.io/webappswg/).
 
-Its mission is to gather and incubate emerging ideas as they pertain to rich text editor development. 
+Its mission is to gather and incubate emerging ideas as they pertain to rich text editor development.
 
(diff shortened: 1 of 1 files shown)


marcoscaceres:
Well, this it's complicated... it's a maintained document, but there is zero expectation of ever standardizing, right?

saschanaz:
@marcoscaceres It's already in that directory, only the link is broken.

johanneswilm:
Hey @gked - I suspect you accidentally moved the execCommand document into the `ActiveDocuments` folder. I assume you meant to put it into the `Archive` folder, right?

marcoscaceres (changes requested):
I should probably go back into archived. We can still maintain it when there is glaring inconsistencies/errors, but let's not give false hope that it's going anywhere.

marcoscaceres:
What might be useful is to figure out what we might be able to salvage and maybe push that into HTML? This may just be tiny little things, but every little thing we salvage is a win. There would have to be clear evidence of interoperability.

johanneswilm:
We decided on the call on Friday to do as you say and put it back into Archived. It does not prevent us from editing the file when need be.

marcoscaceres:
Sounds great @johanneswilm. Thanks for the update!

saschanaz:
> but let's not give false hope that it's going anywhere.

"Archived" also gives false information that those will never be maintained. It should really be "Inactive" or something. (And possibly also renaming `ActiveDocuments` to just `Active`, or even `active`.)

saschanaz:
This is fixed by #264.

PR #251, opened 2020-05-07
Update to a spec draft document

### LINKED ISSUE #176: contenteditable-disabled
Hey, as of TPAC 2017, we have a resolution to create a contenteditable-disabled attribute [1] to be able to keyboard shortcuts and associated menu items from various context menus. It should work like this:
```
<div contenteditable="true" contenteditable-disabled="orderedList formatBold formatFontName">
  ...
</div> 
```
After consulting with developers of Substance.io, CKEditor, ProseMirror and TinyMCE, I would like to propose this list:

```
history // makes the element not be part of any browser-provided undo stack. 
inserHorizontalRule
orderedList
unorderedList
insertLink
insertLineBreak
insertParagraph
formatBold
formatItalic
formatUnderline
formatStrikeThrough
formatSuperscript
formatSubscript
formatJustify
formatDent // indent and outdent
formatSetBlockTextDirection
formatSetInlineTextDirection
formatBackColor
formatFontColor
formatFontName
```

There are some challenges with this:

1. Clipboard actions - one cannot really avoid needing to offer these and accessing the clipboard programmatically is a security issue. So that means one will continue to have the native menu, which can be challenging when combining with an app-provided menu [2]. The solution there seems to be adding the ability to add custom actions to the native menus, but that will need to be dealt with in a separate issue. This was also briefly discussed at TPAC 2017.

2. IE/Firefox by default add controls for tables, positioned items, images, and possibly more. It sounds like Firefox is working on removing these [3]. Should they not be removed, we would need to mention these here as well. 

3. It would be good to have a way to be able to disable all of these in one go. For example `contenteditable-disabled="all"`

[1] https://www.w3.org/2017/11/07-editing-minutes.html#resolution01

[2] https://github.com/ckeditor/ckeditor5/issues/1107 and https://github.com/ProseMirror/prosemirror/issues/7

[3] https://github.com/w3c/editing/issues/171#issuecomment-379143053

gked:
Closes #176 

 * [#250 ] Confirmed there are no ReSpec/BikeShed errors or warnings.

This PR addresses current vision of the shape of contenteditable-disabled proposal.
In previous discussions, the group talked about 

1.  disabling UI that is not available in the editor (e.g. markdown scenarios for font commands)
2.  controlling Undo/Redo UI
3.  disabling some input UI that can show up on a page unexpectedly on user focus. e.g. voice typing UI that Windows team is experimenting with.

We've realized that a more efficient solution for 2 would be to take advantage of [Undo API](https://rniwa.github.io/undo-api/)  to to communicate to UA correct state  of UI at all times.

Edge team has internally discussed 3rd use case and has a number of open questions around if it's the right fit for contenteditable-disabled. Resolution was to spend more time thinking about it.

As a result current draft is only about disabling UI (formatting UI only).


Active Documents/contenteditable-disabled.html
```diff
@@ -0,0 +1,192 @@
...
+    <section id='sotd'>
+      <p></p>
+    </section>
+    <section id='conformance'>
+      <p></p>
+    </section>
+    <section id="introduction">
+        <h2>Introduction</h2>
+        <p>
+            In contenteditable regions of documents, some user agents provide contextual user interface (UI) elements to help accelerate common editing operations.  The UI is meant to enhance the editing experience, but when authors create customized editors that aren’t compatible with user agent UI, it leads to confused users.
```
johanneswilm: aren’t => aren't

johanneswilm (approved):

Active Documents/contenteditable-disabled.html
```diff
@@ -0,0 +1,192 @@
+<!DOCTYPE html>
```
BoCupp-Microsoft: Should this be renamed to match the original file name so we can better diff them?

Active Documents/contenteditable-disabled.html
```diff
@@ -0,0 +1,192 @@
...
+                    company: "Microsoft"
+                },
+                {
+                    name: "Grisha Lyukshin",
+                    mailto: "glyuk@microsoft.com",
+                    company: "Microsoft"
+                }
+            ],
+            github: "https://github.com/w3c/editing",
+            shortName: "contenteditable-interface",
```
BoCupp-Microsoft: Just noticed the short name doesn't match.  Should this be contenteditable-disabled too?

johanneswilm (commented):
Nice work @gked and @BoCupp-Microsoft ! I noticed you removed `supported`. This was something we had added after discussing whether to blacklist or to whitelist to kind of allow whitelisting anyway while giving a preference for blacklisting. 

The reason this is important to JS editors is that it took several of the JS editor companies several months to discover that the Mac touch bar had been launched. During this time, their users would use the buttons of this bar and nothing would happen in the users editors or even worse - the editors were breaking. The same can happen again with new devices launches with new versions of browsers and operating systems that not every development team is aware of immediately. For that reason there was a demand for a way to query all the supported options that then can be disabled subsequently.

Maybe we need to open a ticket about this?

Active Documents/contenteditable-disabled.html
johanneswilm: I'd think so.

Active Documents/contenteditable-disabled.html
johanneswilm: I think that would make it easier to see what exactly has changed, yes.

Active Documents/contenteditable-disabled.html
```diff
@@ -0,0 +1,192 @@
...
+                <li><code>formatremove</code></li>
+                <li><code>formatstrikethrough</code></li>
+                <li><code>formatsubscript</code></li>
+                <li><code>formatsuperscript</code></li>
+                <li><code>formatunderline</code></li>
+            </ul>
+        </p>
+        <p class="issue" title="missing fontSize and insertImage">
+            fontSize / insertImage is missing from beforeInput.  Why?  It seems that it should be listed to create a more complete set.
+            <a href="https://github.com/w3c/editing/issues/249">Issue 249</a>
```
BoCupp-Microsoft: I believe if you use data-number ReSpec will link the issue for you.  I find these features by clicking the ReSpec button, then clicking about ReSpec, and then clicking the link that sounds most like what I want to do.  In this case it took me here: https://github.com/w3c/respec/blob/develop/src/core/issues-notes.js

Active Documents/contenteditable-disabled.html
```diff
@@ -0,0 +1,192 @@
...
+        <p class="note">
+            Note that this specification does not have a shortcut to disable all editing UI.  Authors should consider using a <code>textarea</code> or <code>contenteditable="plaintext-only"</code> for this purpose.
+        </p>
+    </section>
+    <section id="element-contenteditable-mixin">
+        <h2>Extensions to the <dfn>ElementContentEditable</dfn> mixin</h2>
+        <pre class="idl">
+            [Exposed=Window]
+            interface ElementContentEditable {
+                [SameObject, PutForwards=value] readonly attribute DOMTokenList contentEditableDisabled;
```
BoCupp-Microsoft: This readonly might be incorrect if we have a PutForwards.  Need to check the webidl spec.

BoCupp-Microsoft: No actually that looks correct: https://heycam.github.io/webidl/#PutForwards

Active Documents/contenteditable-disabled.html
```diff
@@ -0,0 +1,192 @@
...
+            };
+        </pre>
+        <p>
+            The <code>contentEditableDisabled</code> WebIDL property reflects the contenteditable-disabled DOM content attribute.
```
BoCupp-Microsoft: There are some things in this spec that should be definitions and others that should be linking to those definitions.  Need to determine how to make that happen in ReSpec.  For example, contenteditable-disabled DOM content attribute should probably be linking back to the definition for that phrase.

BoCupp-Microsoft:
The JPG version of ios-editing-buttons looks unused and can be deleted.

Active Documents/contenteditable-disabled.html
gked: sounds good, will do

Active Documents/contenteditable-disabled.html
gked: fixed

Active Documents/contenteditable-disabled.html
gked: yeah, I've created an [issue](https://github.com/w3c/editing/issues/250) to fix DOMTokenList linking, I can use it for all dfns related fixes.

Active Documents/contenteditable-disabled.html
gked: yeah, I've created an issue to fix DOMTokenList linking, I can use it for all dfns related fixes.

Active Documents/contenteditable-disabled.html
gked: > The JPG version of ios-editing-buttons looks unused and can be deleted.

fixed.

Active Documents/contenteditable-disabled.html
gked: > Nice work @gked and @BoCupp-Microsoft ! I noticed you removed `supported`. This was something we had added after discussing whether to blacklist or to whitelist to kind of allow whitelisting anyway while giving a preference for blacklisting.
> 
> The reason this is important to JS editors is that it took several of the JS editor companies several months to discover that the Mac touch bar had been launched. During this time, their users would use the buttons of this bar and nothing would happen in the users editors or even worse - the editors were breaking. The same can happen again with new devices launches with new versions of browsers and operating systems that not every development team is aware of immediately. For that reason there was a demand for a way to query all the supported options that then can be disabled subsequently.
> 
> Maybe we need to open a ticket about this?

Sounds good, I created an [issue ](https://github.com/w3c/editing/issues/252)and tagged with Agenda+ to be discussed tomorrow.

Active Documents/contenteditable-disabled.html
marcoscaceres:
```suggestion
            In contenteditable regions of documents, some user agents provide contextual user interface (UI) elements to help accelerate common editing operations.  The UI is meant to enhance the editing experience, but when authors create customized editors that aren't compatible with user agent UI, it leads to confused users.
```

Active Documents/contenteditable-disabled.html
marcoscaceres: That shouldn't matter... git should have picked up that this was a file rename, not a delete, and done the right thing :(

marcoscaceres: (as a later thing to do later, we need to run HTML 5 tidy or prettier over this doc... will also help with diffing)

Active Documents/contenteditable-disabled.html
```diff
@@ -0,0 +1,192 @@
...
+            ],
+            github: "https://github.com/w3c/editing",
+            shortName: "contenteditable-interface",
+            wg: "Web Applications Working Group",
+            wgURI: "https://www.w3.org/2019/webapps/",
+            wgPatentURI: "https://www.w3.org/2004/01/pp-impl/114929/status",
+            license: "w3c-software-doc"
+        };
+    </script>
+    <script src='https://www.w3.org/Tools/respec/respec-w3c-common' class='remove'></script>
```
marcoscaceres: Nit: common profile is deprecated...

```suggestion
    <script src='https://www.w3.org/Tools/respec/respec-w3c' class='remove'></script>
```

Active Documents/contenteditable-disabled.html
```diff
@@ -0,0 +1,192 @@
...
+</head>
+
+<body>
+    <section id='abstract'>
+        <h2>Abstract</h2>
+        <p>
+            This specification defines an API that indicates to user agents which editing-related commands are applicable to contenteditable regions of a document.  User agents SHOULD take this information into account when presenting any editing-related UI to ensure its applicability.
```
marcoscaceres: Avoid RFC2119 keywords in non-normative sections. 
```suggestion
            This specification defines an API that indicate which editing-related commands are applicable to `contenteditable` regions of a document.  User agents can take this information into account when presenting any editing-related UI to ensure its applicability.
```

"its applicability" is unclear... what is "its" there?

Active Documents/contenteditable-disabled.html
```diff
@@ -0,0 +1,192 @@
...
+        </p>
+        <p>
+            This specification purposefully does not prescribe the means by which a user agent will indicate to the user that some commands are not applicable in a given contenteditable region.
```
marcoscaceres: This sentence reads like a double negative.

Active Documents/contenteditable-disabled.html
```diff
@@ -0,0 +1,192 @@
...
+        </p>
+    </section>
+    <section id='sotd'>
+      <p></p>
```
marcoscaceres:
```suggestion
      <p>This is a work in progress and subject to change without warning.</p>
```

Active Documents/contenteditable-disabled.html
```diff
@@ -0,0 +1,192 @@
...
+    </section>
+    <section id='conformance'>
+      <p></p>
```
marcoscaceres:
```suggestion
```

Active Documents/contenteditable-disabled.html
marcoscaceres: @johanneswilm don't forget you can use GitHub's funky new suggestion button: 

Screenshot 2020-05-08 14 56 31

You can then suggest changes inline that can be merged straight in.

Active Documents/contenteditable-disabled.html
```diff
@@ -0,0 +1,192 @@
...
+    </section>
+    <section id="introduction">
+        <h2>Introduction</h2>
+        <p>
+            In contenteditable regions of documents, some user agents provide contextual user interface (UI) elements to help accelerate common editing operations.  The UI is meant to enhance the editing experience, but when authors create customized editors that aren’t compatible with user agent UI, it leads to confused users.
+        </p>
+        <p>
+            WYSIWYG markdown editors provide a canonical example. While markdown editors typically support bold or italicized text, they often don't support font color or font family customization. System or browser UI elements like those shown for iOS or the Mac Touch Bar below don't have any indication as to which formatting commands are applicable to a particular contenteditable element and should be displayed, and which ones should be disabled or omitted.
+        </p>
+        <figure>
```
marcoscaceres: I don't think we should have this in the spec either... this could change in the future, and it's not appropriate to call out specific implementations in specification. Maybe move all this to a Wiki page here on GitHub.

Active Documents/contenteditable-disabled.html
```diff
@@ -0,0 +1,192 @@
...
+            </figcaption>
+        </figure>
+        <figure>
+            <img src="images/ios-editing-buttons.png" width="300" alt="iOS selection menu showing editing buttons">
+            <figcaption>
+                iOS selection menu showing editing buttons
+            </figcaption>
+        </figure>
+        <p>
+            To address the issue, a new contenteditable-disabled attribute can be applied as shown in the example below.
```
marcoscaceres: 🚨 Whoa!!!! Noob question: is this web compatible??? I though "`-whatever`" was reserved for custom elements? Won't this conflict or cause confusion?

marcoscaceres: Also, isn't this a losing proposition? A content author would need to forever keep updating this list of things to disable, while the UA is constantly trying to enable things. 

Wouldn't an all or nothing approach be better?

Active Documents/contenteditable-disabled.html
```diff
@@ -0,0 +1,192 @@
...
+        </p>
+        <pre class="example">
```
marcoscaceres:
```suggestion
        <pre class="example html">
```

Active Documents/contenteditable-disabled.html
```diff
@@ -0,0 +1,192 @@
...
+                    mailto: "glyuk@microsoft.com",
+                    company: "Microsoft"
+                }
+            ],
+            github: "https://github.com/w3c/editing",
+            shortName: "contenteditable-interface",
+            wg: "Web Applications Working Group",
+            wgURI: "https://www.w3.org/2019/webapps/",
+            wgPatentURI: "https://www.w3.org/2004/01/pp-impl/114929/status",
+            license: "w3c-software-doc"
```
marcoscaceres: We get the right license by default, but we need to enable "xref", for cross references.

```suggestion
            xref: "web-platform", // we will be citing html, infra, and friends.
```

Active Documents/contenteditable-disabled.html
```diff
@@ -0,0 +1,192 @@
...
+        };
+    </script>
+    <script src='https://www.w3.org/Tools/respec/respec-w3c-common' class='remove'></script>
+</head>
+
+<body>
+    <section id='abstract'>
+        <h2>Abstract</h2>
```
marcoscaceres: ReSpec will add this for free... you are welcome 😜
```suggestion
```

Active Documents/contenteditable-disabled.html
marcoscaceres: Note, you should move `<section id='conformance'>` to the end of the document. It's just boring boilerplate.

Active Documents/contenteditable-disabled.html
```diff
@@ -0,0 +1,192 @@
...
+                    &lt;li>strikethrough&lt;/li>
+                    &lt;li>underlines&lt;/li>
+                    &lt;li>horizontal rules&lt;/li>
+                &lt;/ul>
+            &lt;/div>
+        </pre>
+    </section>
+    <section id="contenteditable-disabled">
+        <h2>
+            <code>contenteditable-disabled</code> DOM Content Attribute.
```
marcoscaceres: HTML content attributes should first be approved by the WHATWG folks, and should be really be defined in HTML (the behavior of this stuff can be defined by this spec, however). @annevk and/or @domenic might want to give this a nod of approval. 

We should then make a plan for how to integrate the specs.

Active Documents/contenteditable-disabled.html
```diff
@@ -0,0 +1,192 @@
...
+        </h2>
+        <p>
+            The <dfn><code>contenteditable-disabled</code></dfn> attribute indicates which user agent-supplied editing UI SHOULD be disabled.  If specified, the attribute must have a value that is an <a href=https://html.spec.whatwg.org/multipage/common-microsyntaxes.html#unordered-set-of-unique-space-separated-tokens>unordered
+            set of unique space-separated tokens</a> [[HTML]] which
+            are <a href=https://infra.spec.whatwg.org/#ascii-case-insensitive>ASCII case-insensitive</a> [[INFRA]].
```
marcoscaceres:
```suggestion
            The <code><dfn data-dfn-type="element-attr">contenteditable-disabled</dfn></code> attribute indicates which user agent-supplied editing UI SHOULD be disabled.  When specified, the attribute can contain an [=unordered set of unique space-separated tokens=] that are treated as [=ASCII case-insensitive=].
```

Active Documents/contenteditable-disabled.html
```diff
@@ -0,0 +1,192 @@
...
+        </p>
+        <p>
+            The allowed set of tokens are listed below.  Each token matches the name of an <a href=https://www.w3.org/TR/input-events-1/#events-inputevents>Input Type</a> [[input-events-1]], or is a shortcut that is equivalent to specifying a set tokens.
```
marcoscaceres: Where possible, please alway avoid `<a href>` - worst case, use `data-cite=` instead. 
```suggestion
            The allowed set of tokens are listed below.  Each token matches the name of an <a data-cite="input-events-1#events-inputevents">Input Type</a>, or is a shortcut that is equivalent to specifying a set tokens.
```

Active Documents/contenteditable-disabled.html
```diff
@@ -0,0 +1,192 @@
...
+        </p>
+        <p>
+            When a token is specified, any user agent-supplied editing UI, that when invoked, would result in a beforeinput event having and inputType that matches one of the tokens in the attribute's value, or matches a token that is represented by a shortcut that is included in the attribute's value, SHOULD be disabled, omitted, or otherwise indicated by the user agent to be not applicable.  The exact mechanism for that indication is beyond the scope of this specification.
```
marcoscaceres: The specification is kinda sad that it gives zero assurances about anything 😿 It's SHOULD everywhere, which leads me to believe something is not right with the overall approach.

Active Documents/contenteditable-disabled.html
marcoscaceres: yep, please see also: 
https://github.com/w3c/respec/wiki/data--number

Active Documents/contenteditable-disabled.html
```diff
@@ -0,0 +1,192 @@
...
+                <li><code>formatstrikethrough</code></li>
+                <li><code>formatsubscript</code></li>
+                <li><code>formatsuperscript</code></li>
+                <li><code>formatunderline</code></li>
+            </ul>
+        </p>
+        <p class="issue" title="missing fontSize and insertImage">
+            fontSize / insertImage is missing from beforeInput.  Why?  It seems that it should be listed to create a more complete set.
+            <a href="https://github.com/w3c/editing/issues/249">Issue 249</a>
+        </p>
```
marcoscaceres:
```suggestion
        <div class="issue" data-number="249">
        </div>
```

Active Documents/contenteditable-disabled.html
marcoscaceres:
```suggestion
            The {{ElementContentEditable/contentEditableDisabled}} IDL attribute reflects the contenteditable-disabled DOM content attribute.
```

Active Documents/contenteditable-disabled.html
marcoscaceres: Note: "unordered set of unique space-separated tokens" hans't been exported yet from HTML... still need to do that. In any case, please avoid liking directly to the HTML spec using `<a href=>`, always use `[=whatever you want to link to=]`... and if `[=whatever you want to link to=]` doesn't exist in the [xref database](http://respec.org/xref/), we need to raise an issue on the corresponding spec to get the term exported and added to the database.

marcoscaceres:
Just a note for the editors... the following resource might help you with ReSpec usage: 

https://github.com/w3c/respec/wiki/Shorthands-Guide

marcoscaceres (changes requested):
Couple of things we should sort out... let me know if we should split them into followup bugs or not. 

@saschanaz, you wanna go through this also?

Active Documents/contenteditable-disabled.html
marcoscaceres: When `xref` option is enabled, DOMTokenList will link correctly :)

Active Documents/contenteditable-disabled.html
johanneswilm: I believe the background for this is that it has been mentioned by browser developers that some of the appearance is determined by the OS and that it's not always possible to fully remove certain menus and menu entries. So in those cases they will grey those out in instead or make them visually disabled in some other way.

But using "SHOULD" everywhere also basically means that a browser can be fully compliant with the spec without actually having implemented anything at all, right? In that would it make more sense to write that it "MUST" disable menus in some way without specifying what that way is and those browsers that cannot comply with that then need to be listed as non-compliant?

Active Documents/contenteditable-disabled.html
johanneswilm: @marcoscaceres You mean a way to turn all off and then turn individual ones on again? We have gone back and forth between blacklisting and whitelisting on this. Browser makers generally prefer blacklisting and content authors generally prefer whitelisting. Browser makers are on the long end of the stick, so we went with blacklisting but with the addition that one can query which items are available, so that with a few lines of JavaScript one should effectively be able to implement whitelisting anyway. 

This querying mechanism seems to have been removed with this PR and there is a tuicket here about bringing it back: https://github.com/w3c/editing/issues/252

Active Documents/contenteditable-disabled.html
annevk: This attribute name violates https://w3ctag.github.io/design-principles/#casing-rules.

annevk: And yeah, an issue against HTML with a rough outline of the plan/design would be good.

Active Documents/contenteditable-disabled.html
saschanaz: >That shouldn't matter... git should have picked up that this was a file rename, not a delete, and done the right thing :(

Not when the file changes too much 👀

Active Documents/contenteditable-disabled.html
marcoscaceres: > So in those cases they will grey those out in instead or make them visually disabled in some other way.

Ok, that's interesting. You are saying the functionality gets disabled... so, that could work: It's not so much the presentation/UI aspect that the spec needs to concern itself with, but with the functionality itself... then it can be "MUST" just around the functionality.  

> But using "SHOULD" everywhere also basically means that a browser can be fully compliant with the spec without actually having implemented anything at all, right?

Yeah. 

>  In that would it make more sense to write that it "MUST" disable menus in some way without specifying what that way is and those browsers that cannot comply with that then need to be listed as non-compliant?

So yeah, don't think about this as menus (because then it crosses into areas where browsers compete at the UI level)... just talk about functionality.

Active Documents/contenteditable-disabled.html
BoCupp-Microsoft: The 'its' is referring to editing-related UI in that sentence.  The API provides hints as to what editing operations (and therefore what editing UI) are applicable in a given contenteditable element. We could change it to... "User agents can take this information into account when presenting any editing-related UI to ensure it is applicable to a given contenteditable region."

Active Documents/contenteditable-disabled.html
BoCupp-Microsoft: Maybe... "This specification purposefully does not prescribe the means by which a user agent will indicate to the user the applicability of any editing command.  The user agent could, for example, omit editing UI, disable it, or choose some other appropriate UI treatment."

Active Documents/contenteditable-disabled.html
BoCupp-Microsoft: We could say that user agents MUST NOT display UI that indicates an editing command is applicable if that editing command's corresponding Input Type is present in the value of the contenteditabledisabled attribute?

saschanaz:
This is super hard to read diffs, could we restore the name and rename in a separate PR?

marcoscaceres:
@saschanaz, might be ok to treat this as an entirely new spec and ignore the diff? The spec itself is pretty small and think we’ve identified some showstoppers already.

saschanaz:
> @saschanaz, might be ok to treat this as an entirely new spec and ignore the diff? The spec itself is pretty small and think we’ve identified some showstoppers already.

I think showstoppers don't force abandoning diffs, though.

marcoscaceres:
I guess. Whatever is easiest.

Active Documents/contenteditable-disabled.html
saschanaz: Forgot to push? 👀

gked: hmm, let me check.

gked: I've been pushing my changes through my coworker's fork and the changes are not flowing in, somehow, I am making an attempt to edit the file in the PR.

Active Documents/contenteditable-disabled.html
gked: I pushed another commit. It caused github to disable Commit Suggestion button. I'll make a change on my and push it here.

Active Documents/contenteditable-disabled.html
gked: I pushed another commit. It caused github to disable Commit Suggestion button. I'll make a change on my and push it here.

Active Documents/contenteditable-disabled.html
gked: I pushed another commit. It caused github to disable Commit Suggestion button. I'll make a change on my and push it here.

Active Documents/contenteditable-disabled.html
gked: I am taking Bo's suggestion and will push the changes in the next iteration.

Active Documents/contenteditable-disabled.html
gked: I pushed another commit. It caused github to disable Commit Suggestion button. I'll make a change on my and push it here.

Active Documents/contenteditable-disabled.html
gked: sounds good.

Active Documents/contenteditable-disabled.html
gked: similarly to the issues above, I've changed it on my end.

Active Documents/contenteditable-disabled.html
gked: made changes on my end and will push new changes soon.

Active Documents/contenteditable-disabled.html
gked: @marcoscaceres  and @annevk thank you for your feedback. [Issue](https://github.com/w3c/editing/issues/257) tracking WHATWG discussion initiation. 
I will remove the hyphen. Although, `contenteditabledisabled` seems lengthy to me.

Active Documents/contenteditable-disabled.html
gked: yes, ReSpec now throws the error. I will create an issue once the first change is landed so I can reference updated draft in the issue. [Issue](https://github.com/w3c/editing/issues/258) tracking the work here.

marcoscaceres:
Stupid question... but y u no contenteditable=“disabled”? Is that because any truthy attribute value enables it? Even if so, maybe one can feature detect for it... like set it to “disabled”, if it becomes “”, then remove the attribute. Otherwise, it’s supported. Just a thought.

Active Documents/contentEditableDisabled.html
```diff
@@ -7,84 +7,168 @@
...
             wg: "Web Applications Working Group",
             wgURI: "https://www.w3.org/2019/webapps/",
             wgPatentURI: "https://www.w3.org/2004/01/pp-impl/114929/status",
-            license: "w3c-software-doc"
+            license: "w3c-software-doc",
+            xref: "web-platform" // we will be citing html, infra, and friends.
         };
     </script>
-    <script src='https://www.w3.org/Tools/respec/respec-w3c-common' class='remove'></script>
+     <script src='https://www.w3.org/Tools/respec/respec-w3c' class='remove'></script>
```
saschanaz:
```suggestion
    <script src='https://www.w3.org/Tools/respec/respec-w3c' class='remove'></script>
```

Active Documents/contentEditableDisabled.html
```diff
@@ -7,84 +7,168 @@
...
+                <li><code>formatremove</code></li>
+                <li><code>formatstrikethrough</code></li>
+                <li><code>formatsubscript</code></li>
+                <li><code>formatsuperscript</code></li>
+                <li><code>formatunderline</code></li>
+            </ul>
+        </p>
+        <p class="issue" title="missing fontSize and insertImage">
+            fontSize / insertImage is missing from beforeInput.  Why?  It seems that it should be listed to create a more complete set.
+            <a href="https://github.com/w3c/editing/issues/249">Issue 249</a>
```
saschanaz, line 136: This should reflect the discussion result. Maybe just a note that those are missing.

Active Documents/contentEditableDisabled.html
```diff
@@ -7,84 +7,168 @@
...
+            <ul>
+                <li><code>formatjustify</code>: combines <code>formatjustifycenter</code>, <code>formatjustifyfull</code>, <code>formatjustifyleft</code>, and <code>formatjustifyright</code>.</li>
+                <li><code>formatfont</code>: combines <code>formatfontname</code>, <code>formatfontsize</code>, and <code>formatfontcolor</code></li>
+                <li><code>formatdent</code>: combines <code>formatindent</code> and <code>formatoutdent</code></li>
+                <li><code>insertlist</code>: combines <code>insertorderedlist</code> and <code>insertunorderedlist</code></li>
+                <li><code>formatscript</code>: combines <code>formatsubscript</code> and <code>formatsuperscript</code></li>
+            </ul>
+        </p>
+        <p class="note">
+            Note that this specification does not have a shortcut to disable all editing UI.  Authors should consider using a <code>textarea</code> or <code>contenteditable="plaintext-only"</code> for this purpose.
```
saschanaz, line 154: I think `plaintext-only` shouldn't be recommended as it's currently removed from the spec.

Active Documents/contentEditableDisabled.html
```diff
@@ -7,84 +7,168 @@
...
+        </p>
+    </section>
+    <section id="element-contenteditable-mixin">
         <h2>Extensions to the <dfn>ElementContentEditable</dfn> mixin</h2>
         <pre class="idl">
-           partial interface ElementContentEditable {
-             [SameObject, PutForwards=value] readonly attribute DOMCommandTokenList contentEditableDisabled;
-           };
+            [Exposed=Window]
+            interface ElementContentEditable {
```
saschanaz:
```suggestion
            partial interface mixin ElementContentEditable {
```

Active Documents/contentEditableDisabled.html
```diff
@@ -7,84 +7,168 @@
...
+                <li><code>insertlist</code>: combines <code>insertorderedlist</code> and <code>insertunorderedlist</code></li>
+                <li><code>formatscript</code>: combines <code>formatsubscript</code> and <code>formatsuperscript</code></li>
+            </ul>
+        </p>
+        <p class="note">
+            Note that this specification does not have a shortcut to disable all editing UI.  Authors should consider using a <code>textarea</code> or <code>contenteditable="plaintext-only"</code> for this purpose.
+        </p>
+    </section>
+    <section id="element-contenteditable-mixin">
         <h2>Extensions to the <dfn>ElementContentEditable</dfn> mixin</h2>
```
saschanaz:
```suggestion
        <h2>Extensions to the {{ElementContentEditable}} mixin</h2>
```

saschanaz (changes requested):

Active Documents/contentEditableDisabled.html
johanneswilm: We discussed this at the meeting and agreed to remove them from both specs for now.

gked:
@marcoscaceres and @saschanaz: apologies for leaving this PR hanging. I just got back from a 2 month leave. This PR became really messy in terms of the changes. If you don't mind, I'll abandon it and re-create a new PR with all of the comments here addressed.

saschanaz:
>If you don't mind, I'll abandon it and re-create a new PR with all of the comments here addressed.

Sounds good 👍

johanneswilm:
@gked @BoCupp-Microsoft What is the status here? Can we get these changes merged? The document has been moved since so we'll need to update the PR - or have it all be merged through an alternative PR as you mentioned earlier.

johanneswilm:
@BoCupp-Microsoft @travisleithead I will put this on the agenda for the next call. If it can be resolved before then, we can remove it from the agenda again. Basically, the question is whether there is a new PR being developed somewhere that addresses the feedback that was collected in this PR or whether that is not the case. If it is not the case, should we put it on someone's todo list to create such a PR?

travisleithead:
@johanneswilm yes, it looks like PR https://github.com/w3c/editing/pull/265 is actually the "non-messy" version of this PR that @gked promised. I think it's safe to close this PR unmerged at this point and focus on the one. (I'll transfer your Agenda+ label to that PR so that it can be discussed next week too.

travisleithead:
Closing this PR in favor of https://github.com/w3c/editing/pull/265

PR #259, opened 2020-05-23
fix(execCommand): map outdent to formatOutdent

saschanaz:
An error in #206.


gked:
Apologies for delay on this.
The change looks good.

gked (approved):

PR #264, opened 2020-06-16
move each spec into a separate folder + no graduated/inactive subfolders

johanneswilm:
The participants of the call on 2020-06-12 came up with the resolution to move each spec into a separate folder to make future links be longer lasting. Links to the execCommand document in the HTML specification are to be replaced with a link to the new location. See call log here [1]. This was decided to be the way forward instead of #262.

@BoCupp-Microsoft

[1] https://www.w3.org/2020/06/12-editing-irc#T16-08-53


saschanaz (approved):
I like this, as an inactivity won't change the path. But could you double check we don't have broken links inside the documents?

saschanaz:
So the participants decided to not keep the old paths? I think that was the problem in #261.

johanneswilm:
That is correct. Instead it was decided to make a PR to the HTML spec to change the execCommand links there + try to check whether there are any other links to these documents and then also update those, if possible, and if not possible readd the forward link files.

saschanaz:
Cool 😃

annevk:
Either way you'll need to add redirects. HTML has many snapshots that cannot be updated.

johanneswilm:
@annevk ok, understood. ~~But that then limits the spec links to only those documents that we find links for (so far only execCommand).~~ **EDIT:** Actually it turns out there are links on the web to just about all the linked files.

marcoscaceres (commented):
This is fine, but yeah, let's add the redirects as Anne suggested. At the same time, we should update the HTML spec to make sure links don't break from HTML.

johanneswilm:
OK, I have added the links as suggested. Once this has been merged, the links from the HTML will no longer be broken. After merging this, we can then update the links in the HTML spec to the new (and hopefully permanent) URLs.

marcoscaceres (approved):

PR #265, opened 2020-08-11
updating contenteditabledisabled spec draft

### LINKED ISSUE #176: contenteditable-disabled

gked:
Closes #176 

 * [x] Confirmed there are no ReSpec/BikeShed errors or warnings.

This PR replaces [original PR](https://github.com/w3c/editing/pull/251) that has gotten too messy.


docs/contentEditableDisabled/index.html
```diff
@@ -7,84 +7,164 @@
...
+        </p>
+    </section>
+    <section id="element-contenteditable-mixin">
         <h2>Extensions to the <dfn>ElementContentEditable</dfn> mixin</h2>
         <pre class="idl">
-           partial interface ElementContentEditable {
-             [SameObject, PutForwards=value] readonly attribute DOMCommandTokenList contentEditableDisabled;
-           };
+            [Exposed=Window]
+            interface ElementContentEditable {
```
marcoscaceres, line 163: Ok, this can't go here unfortunately, because it would break HTML. 
https://html.spec.whatwg.org/#contenteditable

This needs to be a PR on HTML itself as part of  the `ElementContentEditable` mixin defined there. 

The values probably need to go into HTML also, but they can point back to our spec, which would defined the expected behavior for each token.

marcoscaceres (dismissed):

docs/contentEditableDisabled/index.html
saschanaz: Wouldn't `partial interface mixin` be enough here?

gked: @marcoscaceres if we move this into HTML spec then I guess, all of it can just live there.
contenteditabledisabled could just be another attribute on conenteditable and we won't need another spec.
@annevk as FYI

Per previous discussion, I created an issue on HTML spec asking for feedback on this: https://github.com/whatwg/html/issues/5863

travisleithead:
WG call: Johannes to file remaining issues (as issues) and then merge this PR so the spec can be read without having to also scan this PR.

travisleithead:
@johanneswilm just a friendly reminder that you have an action item here :)

johanneswilm:
@travisleithead I am aware of that. We are only half-way through the month.

johanneswilm:
@travisleithead This is now ready to be merged. But there is an IPR issue. Grisha was part of the WG when he wrote this code and made the PR. Can we still merge this or do we need him to join the WG as an invited expert?

johanneswilm (approved):

johanneswilm:
As far as I could tell, no parts were missing in this PR and it should be merged as is. I have removed the part of the PR the `supported` attribute, as there was no consensus on removing it. 

There is however the question of whether this should go into the HTML spec, which is we can discuss in issue #331.

travisleithead:
Awesome. Thanks @johanneswilm! Since Grisha's contributions in the PR were covered when he was in the group and working for Microsoft, and since no additional changes have been contributed since he left, and as another Microsoft rep in the group, I'll assume IPR responsibility for this PR on behalf of Microsoft. So that should cover it. I'll override the IPR bot as a special exception on this one.

travisleithead (approved):

travisleithead:
> [travisleithead](https://github.com/travisleithead) marked as non substantive for IPR from ash-nazg.

See my comment above about about Microsoft retaining the IPR responsibility for this PR.

PR #276, opened 2021-01-18
add draft charter for the future Editing WG

siusin:
please review the draft and complete all those tags marked with "todo".

### DIFF
--- a/charter-drafts/editing-2021.html
+++ b/charter-drafts/editing-2021.html
@@ -74,8 +74,7 @@
       <h1 id="title">PROPOSED Editing Working Group Charter</h1>
       <!-- delete PROPOSED after AC review completed -->
 
-      <p class="mission">The <strong>mission</strong> of the <a href="">Editing Working Group</a> is to explore limitations of existing browser primitives, provide use cases for new APIs and suggest solutions either by standardizing of existing behaviors or introducing new APIs, relevant to text editing.
-        The goal is simplifying the creation of fully featured editing systems as well as small editors using JavaScript.</p>
+      <p class="mission">The <strong>mission</strong> of the <a href="">Editing Working Group</a> is to explore limitations in existing browser primitives, provide use cases for new APIs and suggest solutions either by standardizing of existing behaviors or introducing new APIs <strong>relevant for text editing</strong>. The goal is to facilitate the creation of fully-featured editing systems as well as small editors using JavaScript.</p>
 
       <div class="noprint">
         <p class="join"><a href="https://www.w3.org/2004/01/pp-impl/#####/join">Join the Editing Working Group.</a></p>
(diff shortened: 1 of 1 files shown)


johanneswilm:
Thank you for your work on this @siusin !

Changes I would propose:

\- The mission of the Editing Working Group is to explore limitations in existing browser primitives, provide use cases for new APIs and suggest solutions either by standardizing of existing behaviors or introducing new APIs. The goal is to facilitate the creation of fully-featured editing systems as well as small editors using JavaScript..
\+ The mission of the Editing Working Group is to explore limitations in existing browser primitives, provide use cases for new APIs and suggest solutions either by standardizing of existing behaviors or introducing new APIs **relevant for text editing**. The goal is to facilitate the creation of fully-featured editing systems as well as small editors using JavaScript.~.~

\- Teleconferences: topic-specific calls may be held or somthing else
\+ Teleconferences: Not more often than once per month

(Our teleconferences are usually once per month, but we should have the option to skip a meeting in case there is nothing to talk about)

The following features are out of scope, and will not be addressed by this Working group.
\+ * Editing of content other than text and text related content. For example, an image editor, a 3D modeler or a movie cutter would be out of scope.

Normative Specifications
[...]
Input Events level 1
[...]
Input Events level 2
[...]

*Johannes: @siusin We actually discussed at the last teleconference that we will likely be merging these into just one document (dropping level 1, making a few things optional in level 2). We should have a final decision on this within a month. Should we still list both of them here?*

Adopted Draft: [...]
Exclusion Draft: [...]

*Johannes: I don't have access to the W3C online tool for this.*

3. Success criteria
[...]

*Johannes: Does the W3C require us to term it like that we have to get 2 implementations for at least X number of specs? If not, a better formulation in our case would be something along the lines of "The Working Group is successful if it has been able to produce specs and standardize browser behavior so that it is possible to write a basic JS text editor that works across scripts with no more than 3 years development effort." As it doesn't really matter if all our specs are necessary for it or whether the contenteditable and EditContext specs are able to cover all of it.*

[...]

W3C groups 
*Johannes: It is a bit unclear to me what needs to be listed here, but clearly we need to interact with the WebApps WG because that is where the UI events spec lives that we depend upon with the Input Events spec.*
[...]

\- This group primarily conducts its technical work pick one, or both, as appropriate: on the public mailing list public-[email-list]@w3.org (archive) or on GitHub issues.
\+ This group primarily conducts its technical work pick one, or both, as appropriate: on GitHub issues.

[...]

\- A call for consensus (CfC) will be issued for all resolutions (for example, via email, GitHub issue or web-based survey), with a response period from [pick a duration within:] one week to 10 working days
\+ A call for consensus (CfC) will be issued for all resolutions (for example, via email, GitHub issue or web-based survey), with a response period from one week *(Johannes: I don't have a strong opinion on this)*

[...]

Licensing
[...]

*Johannes: I guess we need to pick the license that current documents are under already. But which one is that?*

siusin:
Updated the charter according to [Johannes' comment](https://github.com/w3c/editing/pull/276#issuecomment-762841230) and remove Input Events L1 since the group plan to retire it.

I merged this pull request so that it's easier to review. Please raise an issue if you have any suggestions :)

PR #291, opened 2021-03-24
Fix url/date of latest draft of clipboard API in charter

dontcallmedom:


siusin (approved):
Ahh, good catch. Thank you Dom!

PR #304, opened 2021-05-10
Update EditContext explainer and associated examples

alexkeng:
This PR adds a new section "Difference between DIV with Contenteditable and DIV with EditContext" to the explainer and updates examples/WebIDL sections to reflect the latest status of the EditContext development.

+@gked
+@BoCupp-Microsoft
+@snianu


alexkeng:
-@gked
-@BoCupp-Microsoft
-@snianu

docs/EditContext/explainer.md
```diff
@@ -146,7 +149,7 @@ enum EditContextEnterKeyHint {
     "send"
 };
 
-enum EditContextInputPolicy {
+enum EditContextInputPanelPolicy {
```
gked, line 152: What do we mean by panel here? Besides an on-screen keyboard, is IME candidate window considered to be an input panel? What about inking panel (i think, inking panel would be considered the same as the keyboard panel)?

snianu: I think this is related to virtual keyboard. We should probably change this to `EditContextVirtualKeyboardPolicy `to match our `virtualKeyboardPolicy `attribute proposal

docs/EditContext/explainer.md
```diff
@@ -188,9 +189,81 @@ interface EditContext : EventTarget {
     attribute EventHandler oncompositionend;
 };
 ```
+## Difference between DIV with Contenteditable and DIV with EditContext
```
snianu: Why `<div>` specifically? Why not any contenteditable element?

gked (approved):
we discussed a few questions during the sync and made edits to the PR directly. So, everything looks good on my side.
Thanks!

docs/EditContext/explainer.md
alexkeng: discussed offline. considering inking panel, we'll stick with keyword 'InputPanel' instead of 'VirtualKeyboard' for now.

docs/EditContext/explainer.md
alexkeng: I use <div> just to give a more concrete example so the audience of the explainer can relate and understand the concept more easily. 

(Discussed offline and decided to update the title.)

PR #306, opened 2021-05-17
be more clear about the scope in the charter

siusin:
To clarify:
```
- Is basic manipulation of embedded media (select/delete/move/resize) in scope? We expect that it would be, since "rich text" would include text with inline images.
- Is manipulation of other HTML structures such as tables and lists in scope? Again, we expect that it would also be in scope, but would like to see this clarified.
```

We consider tables and lists as selectors, which have already been listed in [section 1.2](https://w3c.github.io/editing/charter-drafts/editing-2021.html#scope), so no changes made for this question.

/cc @gi11es

### DIFF
--- a/charter-drafts/editing-2021.html
+++ b/charter-drafts/editing-2021.html
@@ -155,6 +155,7 @@
             <li>Clipboard.</li>
             <li>Undo.</li>
             <li>Spellcheck and grammar checking.</li>
+            <li>DOM level manipulation of other elements inside of contenteditable elements.</li>
             <li>Revisiting of under-specified existing editing related features.</li>
             <li>Incubating features related to text editing that may eventually go under the scope of another WG, e.g. <a href="https://www.w3.org/TR/css-highlight-api-1/">Highlight API</a>.</li>
           </ul></p>


johanneswilm (approved):

edg2s:
Hi, I'm the author of the above questions.
> We consider tables and lists as selectors, which have already been listed in section 1.2

Which item in that list are you referring to?

The change here seems sufficiently broad. The place where I had some confusion was what was meant by "text related content" so maybe that could be clarified with examples?

Thanks,

johanneswilm:
> The place where I had some confusion was what was meant by "text related content" so maybe that could be clarified with examples?

That wording there is currently:

"
1.3 Out of Scope
...
* Editing of content other than text and text related content. For example, an image editor, a 3D modeler or a movie cutter would be out of scope."

@edg2s Are you looking for other examples than the three examples we are listing there already? Would you have a suggestion for such an example?

edg2s:
I think I was asking about examples of what _is_ "text related content".

johanneswilm:
@edg2s As far as I can tell, we are only using that term in the "out of scope" section. In the section of what is in-scope, we use other wording. Is any of that wording unclear?

johanneswilm:
The issue here was that we were using the term "editing" to describe what we are working on. And so while the list in the in-scope point describes that quite well, in the out-of-scope section we just wanted to make sure that no-one would get the wrong idea and that we are only referring to "editing" is it is used in a word processor, text editor, text input field, etc., and not "editing" of other things such as images, movies, etc. . 

So for example if someone would get the idea that we should create a movie editor or a sound editor and just put it inside of a contenteditable element so that it would fall into scope of our charter, then the out-of-scope section would make sure that it would be abundantly clear that this is not the right working group to work on that.

edg2s:
Yeah - that part was clear. I understand you don't necessarily want to go into detail in the "out of scope" section.

I think I'm still not clear on where tables and lists are covered then. You state that they are "selectors" and therefore in section 1.2, but I don't see that term in that section. (Or did you mean selection?)

Also what is the current thinking on tables and lists? Would one expect creation of these elements to be covered? Manipulation (indent/de-indent, add/remove cells/rows/columns)?

johanneswilm:
> I think I'm still not clear on where tables and lists are covered then. You state that they are "selectors" and therefore in section 1.2, but I don't see that term in that section. (Or did you mean selection?)

I am not sure what was meant with that, but I think lists and tables and bold tags and all the elements used within text editing are quite clearly covered under "DOM level manipulation of other elements inside of contenteditable elements."

It would also not be consistent with what the Editing Taskforce has been working on over the past 5 years to now exclude lists and tables.

johanneswilm:
@edg2s Is there a particular issue you would like to take up? We have been using the github repository to get some clarification on how some low level execCommand related things work in the past whenever it was important and I would think we will continue to do that. But we don't list execCommand as a deliverable as really finishing it entirely and standardize all behavior across browsers would likely be too large a job for us.

As for tables - as you will know, Firefox used to show handlers by default. One of the largest changes there has been to turn them off by default so that Firefox works like the other browsers by default. Doing good selection in all directions of table cells requires having multiple selection ranges, so it seems most editors have programmed their own JavaScript solution for that part and I am not aware of much of a need to get this handled by browsers. But if there is a request for writing a spec that deals more with lists and tables in a word processor like environment - then yes, this would be the right space to discuss it.

edg2s:
> I am not sure what was meant with that, but I think lists and tables and bold tags and all the elements used within text editing are quite clearly covered under "DOM level manipulation of other elements inside of contenteditable elements."
> 
> It would also not be consistent with what the Editing Taskforce has been working on over the past 5 years to now exclude lists and tables.

Thanks, that's good to know.

edg2s:
> ... most editors have programmed their own JavaScript solution for that part and I am not aware of much of a need to get this handled by browsers. But if there is a request for writing a spec that deals more with lists and tables in a word processor like environment - then yes, this would be the right space to discuss it.

This is all useful context as well, thanks. I'm didn't have a specific proposal in mind, but good to know what is up for discussion.

gked (approved):

PR #309, opened 2021-06-22
VirtualKeyboardAPI  spec draft

gked:
This is the first draft of the spec to iterate on.
No tests, so far, just the wording around the behavior of the API.
There are a few respec errors that we'll be looking at as well.

Implementation commitment:

 * [ ] WebKit ()
 * [X] Chromium (https://bugs.chromium.org/p/chromium/issues/detail?id=856269)
 * [ ] Gecko ()


docs/virtualkeyboard/index.html
```diff
@@ -0,0 +1,455 @@
...
+            </dt>
+              <dd>
+                <p>
+                  The method must follow these steps:
+                </p>
+                <ol>
+                  <li>If the <a>document</a> doesn't have <a href="https://html.spec.whatwg.org/multipage/interaction.html#sticky-activation">sticky user activation</a>, abort these steps.
+
+                  </li>
+                  <li>Trigger {{VirtualKeyboard/show()}} call that invokes a system API to show the VK.
```
yoavweiss: That seems kinda circular. It'd be good to define what the call actually does.

docs/virtualkeyboard/index.html
```diff
@@ -0,0 +1,455 @@
...
+        <dl>
+            <dt>
+                <dfn>show()</dfn> method
+            </dt>
+              <dd>
+                <p>
+                  The method must follow these steps:
+                </p>
+                <ol>
+                  <li>If the <a>document</a> doesn't have <a href="https://html.spec.whatwg.org/multipage/interaction.html#sticky-activation">sticky user activation</a>, abort these steps.
```
yoavweiss: If I'm reading the sticky activation algorithm correctly, it applies to a Window object, not a Document.

yoavweiss: It would be good to clarify which Document (or Window) we're talking about

docs/virtualkeyboard/index.html
```diff
@@ -0,0 +1,455 @@
...
+                <dfn>hide()</dfn> method
+            </dt>
+              <dd>
+                <p>
+                  The method must follow these steps:
+                </p>
+                <ol>
+                  <li>If the <a>document</a> doesn't have <a href="https://html.spec.whatwg.org/multipage/interaction.html#sticky-activation">sticky user activation</a>, abort these steps.
+                  </li>
+                  <li>Trigger {{VirtualKeyboard/hide()}}call that invokes a system API to hide the keyboard.
```
yoavweiss: Same comments as above

docs/virtualkeyboard/index.html
```diff
@@ -0,0 +1,455 @@
...
+                  </li>
+                </ol>
+               </dd>
+               <p class="note">
+                Platform heuristics may impose additional restrictions on VK {{VirtualKeyboard/show()}} and {{VirtualKeyboard/hide()}}.
+              </p>
+              <p class="note">
+                A few people have expressed concerns around {{VirtualKeyboard/show()}} and {{VirtualKeyboard/hide()}} methods not being promise-based. However, we couldn't find any real use cases around needing promises in this type of operation.
```
yoavweiss: Examples below are showing them returning a Promise. Would be good to clarify which variant is implemented.

docs/virtualkeyboard/index.html
yoavweiss: For example, I'm guessing it's impacting `boundingRect` in some way.

docs/virtualkeyboard/index.html
```diff
@@ -0,0 +1,455 @@
...
+              </p>
+            <dt>
+              <dfn>overlaysContent</dfn>
+            </dt>
+            <dd>
+              <p>
+                The attribute enables the author to opt-out of resizing the visual viewport.
```
yoavweiss: Would be good to define a setter that impacts the processing and does the actual opt-out.

docs/virtualkeyboard/index.html
```diff
@@ -0,0 +1,455 @@
...
+              </p>
+            </dd>
+            <dt>
+              <dfn>boundingRect</dfn>
+            </dt>
+            <dd>
+            <p>
+                The attribute reports the intersection of the VK with the layout viewport in client coordinates.
```
yoavweiss, line 178: You should define a getter that defines what's returned here

docs/virtualkeyboard/index.html
yoavweiss: Similarly, it probably fires the `geometrychange` event at some point.

yoavweiss: It might be best to define the processing of what happens when the virtual keyboard is displayed and folded and then hook into that processing from these explicit methods, but also from places where the UA can decide to do that automatically (e.g. from editable content when it's getting focus)

docs/virtualkeyboard/index.html
```diff
@@ -0,0 +1,455 @@
...
+        <h2>Extensions to the <dfn>ElementContentEditable</dfn> mixin</h2>
+        <pre class="idl">
+           partial interface ElementContentEditable {
+            [CEReactions, Reflect, ReflectOnly=("auto","manual")] 
+            attribute DOMString virtualKeyboardPolicy;
+           };
+        </pre>
+        <p>The <dfn><code>virtualKeyboardPolicy</code></dfn> is an <a href=https://html.spec.whatwg.org/multipage/common-microsyntaxes.html#enumerated-attribute>enumerated attribute</a> whose keywords are the empty string, auto, and manual.
+        <p>When specified on an element that is a contenteditable host, <dfn><code>auto</code></dfn> causes the corresponding editable element to automatically show the VK when it is focused or tapped & <dfn><code>manual</code></dfn> 
+            decouples focus and tap on the editable element from changes in the VK’s current state - the VK remains as it was.
```
yoavweiss, line 203: Similarly to the above, you probably want to define the processing for this attribute: what happens when it's set, changed, removed, etc.

docs/virtualkeyboard/index.html
```diff
@@ -0,0 +1,455 @@
...
+        </p>
+        <section>
+            <h3>Keyboard inset variables</h3>
+            <table data-for="env()">
+                <tr>
+                    <th>Name
+                    <th>Value
+                <tr>
+                    <td><dfn>keyboard-inset-top</dfn>
+                    <td><a href="https://drafts.csswg.org/css-values-3/#length-valuehttps://drafts.csswg.org/css-values-3/#length-value"> length</a>
```
yoavweiss, line 241: You probably want to define the semantics of each of these variables and what its value is set to.

gked: updated

docs/virtualkeyboard/index.html
gked: updated

docs/virtualkeyboard/index.html
gked: updated

docs/virtualkeyboard/index.html
gked: updated

docs/virtualkeyboard/index.html
gked: updated

docs/virtualkeyboard/index.html
gked: updated

docs/virtualkeyboard/index.html
gked: updated

docs/virtualkeyboard/index.html
gked: updated

gked:
@yoavweiss We addressed your feedback. Could you please let us know if they resolve your questions? One thing, I am trying to resolve currently is making adding the ability to make a PR readable with .pr-preview.json file.

docs/virtualkeyboard/index.html
```diff
@@ -0,0 +1,492 @@
...
+          , "geometry"]
+        };
+    </script>
+    <script src='https://www.w3.org/Tools/respec/respec-w3c' class='remove'></script>
+</head>
+
+<body>
+    <section id='abstract'>
+        <p>
+            The {{VirtualKeyboard}} API provides web developers a more granular control of on-screen keyboard (virtual keyboard or VK). The API also provides information about VK geometry, such as rect occlusion and gives ability to opt-out from resizing of the visual viewport in the browser.
```
BoCupp-Microsoft: nit: grammar issue with "a more granular control"

Suggested new abstract: The {{VirtualKeyboard}} API provides authors with greater control over the visibility of the virtual keyboard (VK), and greater ability to adapt the layout of web pages when VK visibility changes.

docs/virtualkeyboard/index.html
```diff
@@ -0,0 +1,492 @@
...
+    </section>
+    <section id='sotd'>
+      <p>
+      </p>
+    </section>
+    <section>
+        <h2>Introduction</h2>
+        <P>This section is non-normative.</P>
+        <p>
+            The Virtual Keyboard (VK) is the on-screen keyboard used for input in scenarios where a hardware keyboard may not be available. Unlike hardware keyboard, VK can adapt its shape to optimize for the expected type of input.
```
BoCupp-Microsoft: Suggest cutting this sentence as it doesn't seem relevant to the API:  Unlike hardware keyboard, VK can adapt its shape to optimize for the expected type of input.

I think it was useful in the explainer since we were pointing out a gap that inputMode was overloaded to control both shape and visibility and couldn't do both at once, but in the spec I don't see that it's necessary.

docs/virtualkeyboard/index.html
```diff
@@ -0,0 +1,492 @@
...
+            In addition, user agents respond to the presence of the VK, without any exposure of this information to web developers. This document proposes a new web API surface for VK that developers can use to enable better customization of their webpage's content and experiences.
```
BoCupp-Microsoft: I think it would be useful in this paragraph to expand on how user agents currently respond to the presence of the VK and what alternatives this API offers.  As is I don't think it adds anything compared to the abstract.

Maybe:  User agents respond to the presence of the VK by:
1. Repositioning the user agent above the VK
2. Reducing the size of the layout viewport so the VK doesn't occlude it
3. Reducing the size of the visual viewport and padding the layout viewport to ensure it can be shifted above the VK

This API provides a fourth option that allows the user agent to leave its layout and visual viewports unchanged and instead provide information about the intersection of the VK and layout viewport so that the author can adapt the layout of their web pages using JavaScript or CSS environment variables.

It would be good to have graphics illustrating the difference between these behaviors.

docs/virtualkeyboard/index.html
```diff
@@ -0,0 +1,492 @@
...
+    <section data-dfn-for="VirtualKeyboard">
+        <h2>The <dfn>VirtualKeyboard</dfn> Interface</h2>
+        <pre class="idl">
+            partial interface Navigator {
+                [SameObject] readonly attribute VirtualKeyboard virtualKeyboard;
+            };
+            
+            [Exposed=Window]
+            interface VirtualKeyboard : EventTarget {
+                undefined show();
```
BoCupp-Microsoft, line 106: I think this should be `void` instead of `undefined`.

https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/modules/virtualkeyboard/virtual_keyboard.idl;l=1?q=virtual_keyboard.idl&sq=&ss=chromium

docs/virtualkeyboard/index.html
```diff
@@ -0,0 +1,492 @@
...
+            </dt>
+              <dd>
+                <p>
+                  The method must follow these steps:
+                </p>
+                <ol>
+                    <li>If the browsing context's <a href="https://html.spec.whatwg.org/multipage/browsers.html#active-window">active window</a> does not have <a href="https://html.spec.whatwg.org/multipage/interaction.html#sticky-activation">sticky user activation</a>, abort these steps.
+                  </li>
+                  <li>
+                    If the focused element's <a href="https://html.spec.whatwg.org/multipage/interaction.html#attr-contenteditable">contenteditable</a> value is false, then abort these steps.
```
BoCupp-Microsoft: I don't think this is meant to be limited to contenteditable.  It should say something like: If the focused (is focused the right word) element is not editable...

We want to include textareas, text-like inputs, and any inputs into which the user can type (in the near future EditContext associated elements too).

docs/virtualkeyboard/index.html
```diff
@@ -0,0 +1,492 @@
...
+                  </li>
+                  <li>
+                    If the <code>virtualKeyboardPolicy</code> is not <code></code>manual</code> or <a href="https://html.spec.whatwg.org/multipage/interaction.html#input-modalities:-the-inputmode-attribute">inputMode</a> attribute value is <a href="https://html.spec.whatwg.org/multipage/interaction.html#attr-inputmode-keyword-none">none</a> then abort these steps.
```
BoCupp-Microsoft: nit: "inputMode</a> attribute value" should be inputMode</a>'s attribute value

docs/virtualkeyboard/index.html
```diff
@@ -0,0 +1,492 @@
...
+                  </li>
+                  <li>
+                      Call the system API to show the VK.
```
BoCupp-Microsoft, line 131: I think there should be a note somewhere near here that says additional OS heuristics may prevent the VK from being shown.  Maybe the "Call the system API to show the VK." should instead say something like "Request that the OS show the VK."

docs/virtualkeyboard/index.html
```diff
@@ -0,0 +1,492 @@
...
+                  </li>
+                  <li>
+                    When the VK is shown by the system, {{VirtualKeyboard/ongeometrychange()}} event is fired that contains the intersection of the keyboard geometry and layout viewport in client coordinates.
+                  </li>
+                </ol>
```
BoCupp-Microsoft, line 136: I think you could not have this bullet here.  I think it should be covered below when you define when the `geometrychange` event fires.

docs/virtualkeyboard/index.html
```diff
@@ -0,0 +1,492 @@
...
+               </dd>
+            <dt>
+                <dfn>hide()</dfn> method
```
BoCupp-Microsoft, line 139: most comments above for `show()` also apply to the steps below for `hide()`

docs/virtualkeyboard/index.html
```diff
@@ -0,0 +1,492 @@
...
+                <li>
+                    Call the system API to hide the VK.
+                </li>
+                <li>
+                    When the VK is hidden by the system, {{VirtualKeyboard/ongeometrychange()}} event is fired that contains the intersection of the keyboard geometry and layout viewport in client coordinates.
+                </li>
+                </ol>
+               </dd>
+               <p class="note">
+                Platform heuristics may impose additional restrictions on VK {{VirtualKeyboard/show()}} and {{VirtualKeyboard/hide()}}. e.g., On Windows the pointer type has to be
```
BoCupp-Microsoft: nit: On should not be capitalized

docs/virtualkeyboard/index.html
```diff
@@ -0,0 +1,492 @@
...
+                touch or pen.
```
BoCupp-Microsoft: nit: other paragraphs don't have hard line breaks dividing their content... remove this one?

docs/virtualkeyboard/index.html
```diff
@@ -0,0 +1,492 @@
...
+              </p>
+              <p class="note">
+                A few people have expressed concerns around {{VirtualKeyboard/show()}} and {{VirtualKeyboard/hide()}} methods not being promise-based. However, we could not find any real use cases around needing promises in this type of operation.
```
BoCupp-Microsoft: Can you link to an issue?

docs/virtualkeyboard/index.html
```diff
@@ -0,0 +1,492 @@
...
+              </p>
+            <dt>
+              <dfn>overlaysContent</dfn>
+            </dt>
+            <dd>
+              <p>
+                To set the attribute to opt-out of resizing the visual viewport,
```
BoCupp-Microsoft: Can you say something slightly stronger here: a user agent MUST NOT resize its layout viewport or visual viewport when overlaysContent is true?

Also, I think only top-level browsing contexts are allowed to set this, right?  This would probably be the place to mention requirements like that.  I think there are other requirements too (secure origin maybe)?

docs/virtualkeyboard/index.html
```diff
@@ -0,0 +1,492 @@
...
+                                <script type="module">
+                                    navigator.virtualKeyboard.overlaysContent = true
+                                </script>
+                            </xmp>
+                        </pre>
+                        </aside>
+            </section>
+    </section>
+    <section>
+        <h2>Privacy and Security Considerations</h2>
```
BoCupp-Microsoft, line 461: Several pieces of information in this section seem like a set of claims that this API doesn't come with significant risk to privacy or security.  I don't think we need to cover that.  I expected this section to enumerate important implementation details that UAs may or must follow to ensure a secure and privacy preserving implementation.  Additionally, I think it would be appropriate to enumerate the responsibilities of authors, if any, to ensure they don't create privacy or security issues when using the API.  

Maybe [IndexedDB](https://www.w3.org/TR/IndexedDB-2/#privacy) is a good example.

docs/virtualkeyboard/index.html
```diff
@@ -0,0 +1,492 @@
...
+            </table>
+            
+            <p>The keyboard insets are six <a href="https://drafts.csswg.org/css-env-1/#environment">environment variables</a> that define a rectangle by
+            its top, right, bottom, and left insets from the edge of the viewport. Default value of the keyboard insets are "0px" if a fallback value is not provided else it gets updated when {{VirtualKeyboard/boundingRect}} value changes.
+            The width & height insets are calculated from the remaining insets for developer ergonomics.
+            </p>
+        </section>
+    </section>
+    <section>
+        <h2>Examples</h2>
```
BoCupp-Microsoft, line 266: I think there were [good images in the explainer](https://github.com/MicrosoftEdge/MSEdgeExplainers/blob/main/VirtualKeyboardAPI/explainer.md#examples) to go with these examples.  Can you include them to help illustrate the impact of the code?

docs/virtualkeyboard/index.html
gked: updated

docs/virtualkeyboard/index.html
gked: agreed - updated.

docs/virtualkeyboard/index.html
gked: updated

docs/virtualkeyboard/index.html
gked: we were getting respec errors when we used `void`. Web IDL recommends using `undefined` instead of void https://heycam.github.io/webidl/#idl-undefined

docs/virtualkeyboard/index.html
gked: updated.

docs/virtualkeyboard/index.html
gked: updated

docs/virtualkeyboard/index.html
gked: We added the note down below. it's going to break the flow of the algorithm if we add it here.

docs/virtualkeyboard/index.html
gked: reworded the step slightly. Let us know if this works.

docs/virtualkeyboard/index.html
gked: noted.
FYI - @snianu  and I realized that we should not not impose editability requirement to close the hide the VK.

docs/virtualkeyboard/index.html
gked: updated

docs/virtualkeyboard/index.html
gked: updated

docs/virtualkeyboard/index.html
gked: updated

docs/virtualkeyboard/index.html
gked: Updated the spec to reflect on your first comment.

Is defining SecureContext in web idl enough so we don't have to mention it here?

docs/virtualkeyboard/index.html
gked: added

.pr-preview.json
```diff
@@ -1,4 +1,4 @@
 {
-    "src_file": "index.html",
+    "src_file": "\\docs\\virtualkeyboard\\index.html",
```
johanneswilm: are those Windows-slashes?

gked: Thanks! fixed. This is probably why the preview functionality didn't work :)

docs/virtualkeyboard/index.html
gked: Updated. Is this what you had in mind?

BoCupp-Microsoft (approved):
This looks good enough to me to merge and subsequent issues we can deal with in smaller easier to review PRs.

PR #311, opened 2021-06-28
Remove duplicate pickling explainer sections

dway123:
There previously were duplicate "Risks" and "Security and Privacy" sections in the pickling explainer. Remove the version with less up-to-date language.

(In this PR, the section we're keeping is right above the section we're deleting)


snianu (approved):
Thanks!

Issue #314, opened 2021-07-07
VirtualKeyboard spec feedback


yoavweiss:
Looking at https://w3c.github.io/editing/docs/virtualkeyboard/, there are a few points that can be made clearer:
* In [section 3](https://w3c.github.io/editing/docs/virtualkeyboard/#the-virtualkeyboard-interface), "If the browsing context's active window" is inaccurate. It'd be better IMO to replace it with "[this](https://heycam.github.io/webidl/#this)'s [relevant global](https://html.spec.whatwg.org/#concept-relevant-global)" while asserting that it's indeed a Window (which it should be, as the method is only exposed on a Window).
* I'm assuming that step 5 in both `show()` and `hide()` should be run in parallel, and the event doesn't fire synchronously. You should make that explicit. Depending if step 4 takes time, you may also want to run it in parallel as well.
* `overlaysContent` is missing a setter that would impact processing of the virtual keyboard and its interaction with the content. As is, it's really unclear what it's supposed to be doing. (e.g. I'm assuming you didn't mean for it to impact viewport changes that are unrelated to virtual keyboards)
* `boundingRect` should have a clear getter that defines what the returned values are.

/cc @gked

snianu:
@yoavweiss Addressed all your comments [here](https://github.com/w3c/editing/pull/317)

snianu:
This issue has been addressed in the linked PR.

PR #317, opened 2021-07-23
Address VK spec review comments.

snianu:
https://github.com/w3c/editing/issues/314

Closes #????

 * [X] Confirmed there are no ReSpec/BikeShed errors or warnings.

Implementation commitment:

 * [X] Chromium (https://bugs.chromium.org/p/chromium/issues/detail?id=856269)

### DIFF
--- a/docs/virtualkeyboard/index.html
+++ b/docs/virtualkeyboard/index.html
@@ -115,7 +115,7 @@
         </p>
         <dfn>overlaysContent</dfn>
         <p>
-            A boolean, initially `false`. When this attribute is set to `true`, a user agent MUST NOT resize its layout viewport or <a href="https://wicg.github.io/visual-viewport/#dom-visualviewport">visual viewport</a>.
+            A boolean, initially `false`. When this attribute is set to `true`, a user agent MUST NOT resize its [=document=]'s <a href="https://drafts.csswg.org/css2/#viewport">viewport</a> or <a href="https://wicg.github.io/visual-viewport/#dom-visualviewport">visual viewport</a>.
         </p>
         <dl>
             <dt>
@@ -127,23 +127,28 @@
                 </p>
                 <ol>
                   <li>
-                    Let |window| be [=this=]'s <a href="https://html.spec.whatwg.org/#concept-relevant-global">relevant global object</a>. If |window| is not a <a href="https://html.spec.whatwg.org/multipage/window-object.html#window">Window object</a>, then abort these steps.
+                    Let |window| be [=this=]'s [=relevant global object=]. Assert that |window| is a {{Window}} object.
                   </li>
                   <li>
-                    If |window| does not have <a href="https://html.spec.whatwg.org/multipage/interaction.html#sticky-activation">sticky user activation</a>, abort these steps.
+                    If |window| does not have [=sticky activation=]</a>, abort these steps.
                   </li>
                   <li>
-                    If the focused element is not a form control (such as the value of <a href="https://html.spec.whatwg.org/multipage/form-elements.html#the-textarea-element">textarea</a> elements), or an <a href="https://html.spec.whatwg.org/multipage/interaction.html#editing-host">editing host</a>  (e.g., using <a href="https://html.spec.whatwg.org/multipage/interaction.html#attr-contenteditable">contenteditable</a>) then abort these steps.
+                    If the focused element is not a form control (such as the value of the [=textarea=] element), or an [=editing host=]</a>  (e.g., using <a href="https://html.spec.whatwg.org/multipage/interaction.html#attr-contenteditable">contenteditable</a>) then abort these steps.
                   </li>
                   <li>
-                    If the <code>virtualKeyboardPolicy</code> is not <code></code>manual</code> or <a href="https://html.spec.whatwg.org/multipage/interaction.html#input-modalities:-the-inputmode-attribute">inputMode's</a> attribute value is <a href="https://html.spec.whatwg.org/multipage/interaction.html#attr-inputmode-keyword-none">none</a> then abort these steps.
+                    If the {{ElementContentEditable/virtualKeyboardPolicy}} is not {{manual}} or <a href="https://html.spec.whatwg.org/multipage/interaction.html#input-modalities:-the-inputmode-attribute">inputMode's</a> attribute value is <a href="https://html.spec.whatwg.org/multipage/interaction.html#attr-inputmode-keyword-none">none</a> then abort these steps.
                   </li>
                   <li>
                       Call the system API to show the VK.
                   </li>
                   <li>
-                    When the VK is shown by the system, <a href="https://html.spec.whatwg.org/#in-parallel">in parallel</a> the OS reports the bounding rectangle of the keyboard. Update the {{VirtualKeyboard/boundingRect}} which is the intersection of the keyboard rectangle reported by the OS and the layout viewport. Fire {{VirtualKeyboard/ongeometrychange()}} event to JS.
-                  </li>
+                    [=In parallel=], follow these steps:
+                    <ol>
+                        <li>Wait for the virtual keyboard to be shown by the system.</li>
+                        <li>Call {{VirtualKeyboard/set the virtual keyboard bounding rect}} with the keyboard's OS reported bounding rectangle and the [=document=]'s <a href="https://drafts.csswg.org/css2/#viewport">viewport</a> rectangle.</li>
+                        <li>[=Fire an event=] named "geometrychange" at [=this=].</li>
+                    </ol>
+                   </li>
                 </ol>
                </dd>
             <dt>
@@ -155,24 +160,29 @@
                 </p>
                 <ol>
                   <li>
-                    Let |window| be [=this=]'s <a href="https://html.spec.whatwg.org/#concept-relevant-global">relevant global object</a>. If |window| is not a <a href="https://html.spec.whatwg.org/multipage/window-object.html#window">Window object</a>, then abort these steps.
+                    Let |window| be [=this=]'s [=relevant global object=]. Assert that |window| is a {{Window}} object.
                   </li>
                   <li>
-                    If |window| does not have <a href="https://html.spec.whatwg.org/multipage/interaction.html#sticky-activation">sticky user activation</a>, abort these steps.
+                    If |window| does not have [=sticky activation=]</a>, abort these steps.
                   </li>
                   <li>
-                    If the focused element's <code>virtualKeyboardPolicy</code> is not <code>manual</code> or <a href="https://html.spec.whatwg.org/multipage/interaction.html#input-modalities:-the-inputmode-attribute">inputMode's</a> attribute value is <a href="https://html.spec.whatwg.org/multipage/interaction.html#attr-inputmode-keyword-none">none</a> then abort these steps.
+                    If the focused element's {{ElementContentEditable/virtualKeyboardPolicy}} is not {{manual}} or <a href="https://html.spec.whatwg.org/multipage/interaction.html#input-modalities:-the-inputmode-attribute">inputMode's</a> attribute value is <a href="https://html.spec.whatwg.org/multipage/interaction.html#attr-inputmode-keyword-none">none</a> then abort these steps.
                   </li>
                   <li>
                     Call the system API to hide the VK.
                   </li>
                   <li>
-                    When the VK is hidden by the system, <a href="https://html.spec.whatwg.org/#in-parallel">in parallel</a> the OS reports the bounding rectangle of the keyboard (which has all 0 values). Create a {{DOMRect}} with 0 values and return it when {{VirtualKeyboard/boundingRect}} is queried. Fire {{VirtualKeyboard/ongeometrychange()}} event to JS.
-                  </li>
+                    [=In parallel=], follow these steps:
+                    <ol>
+                        <li>Wait for the virtual keyboard to be hidden by the system.</li>
+                        <li>Call {{VirtualKeyboard/set the virtual keyboard bounding rect}} with the keyboard's OS reported bounding rectangle (which has all 0 values) and the [=document=]'s <a href="https://drafts.csswg.org/css2/#viewport">viewport</a> rectangle.</li>
+                        <li>[=Fire an event=] named "geometrychange" at [=this=].</li>
+                    </ol>
+                   </li>
                 </ol>
                </dd>
                <p class="note">
-                Platform heuristics may impose additional restrictions on VK {{VirtualKeyboard/show()}} and {{VirtualKeyboard/hide()}}. e.g., on Windows the pointer type has to be touch or pen.
+                Platform heuristics may impose additional restrictions on VK {{VirtualKeyboard/show()}} and {{VirtualKeyboard/hide()}}. e.g., on Windows the pointer type MUST be touch or pen.
               </p>
               <p class="note">
                 A few people have <a href="https://github.com/MicrosoftEdge/MSEdgeExplainers/issues/415">expressed concerns</a> around {{VirtualKeyboard/show()}} and {{VirtualKeyboard/hide()}} methods not being promise-based, but we believe it is a better design for web developers to use {{VirtualKeyboard/ongeometrychange}} event since it is fired when VK visibility changes, making the return values unnecessary.
@@ -193,24 +203,34 @@
             </dt>
             <dd>
             <p>
-                The attribute reports the intersection of the VK with the layout viewport in client coordinates. It is set when {{VirtualKeyboard/ongeometrychange}} event is fired. To create a {{VirtualKeyboard/boundingRect}} follow these steps:
-                <ol>
-                    <li>
-                        Let `osk` be the on-screen keyboard rectangle in CSS pixels that is provided by the OS when the VK visibility changes.
-                    </li>
-                    <li>
-                        Let `lv` be the layout viewport's bounds in CSS pixels.
-                    </li>
-                    <li>
-                        Let `bounds` be a {{DOMRect}} object.
-                    </li>
-                    <li>
-                        Set `bounds` to be the intersection of `lv` and `osk` in CSS pixels.
-                    </li>
-                    <li>
-                        Return `bounds`.
-                    </li>
-                </ol>
+                The attribute reports the intersection of the VK with the [=document=]'s <a href="https://drafts.csswg.org/css2/#viewport">viewport</a> in client coordinates. Call {{VirtualKeyboard/set the virtual keyboard bounding rect}}.
+                <p>
+                    <p><dfn>set the virtual keyboard bounding rect</dfn></p>
+                    To compute the intersection of VK with the [=document=]'s <a href="https://drafts.csswg.org/css2/#viewport">viewport</a>, run these steps:
+                    <ol>
+                        <li>
+                            Let |osk| be the on-screen keyboard rectangle that is the result of running the algorithm in {{VirtualKeyboard/show()}} or {{VirtualKeyboard/hide()}} on [=this=].
+                        </li>
+                        <li>
+                            Let |lv| be the [=document=]'s <a href="https://drafts.csswg.org/css2/#viewport">viewport that is the result of running the algorithm in {{VirtualKeyboard/show()}} or {{VirtualKeyboard/hide()}} on [=this=].
+                        </li>
+                        <li>
+                            Map |osk| to the coordinate space of |lv|.
+                        </li>
+                        <li>
+                            Let |bounds| be a {{DOMRect}} object.
+                        </li>
+                        <li>
+                            Update |bounds| by intersecting |lv| with the |osk|.
+                        </li>
+                        <li>
+                            Map |bounds| to the coordinate space of |lv|.
+                        </li>
+                        <li>
+                            Return |bounds|.
+                        </li>
+                    </ol>
+               </p>
             </p>
             <p>
                 The {{VirtualKeyboard/boundingRect}} getter steps are to return [=this=]'s </a>{{VirtualKeyboard/boundingRect}}.
@@ -221,7 +241,7 @@
               </dt>
               <dd>
               <p>
-                  The event is dispatched when the intersection of the VK and the layout viewport changes, e.g., in response to the VK being shown or hidden or the browser window being repositioned.
+                  The event is dispatched when the intersection of the VK and the [=document=]'s <a href="https://drafts.csswg.org/css2/#viewport">viewport</a> changes, e.g., in response to the VK being shown or hidden or the browser window being repositioned.
               </p>
               </dd>
         </dl>
@@ -349,7 +369,7 @@
                 </aside>
             </section>
             <section>
-                <h3>Examples with usage of geometry and ovelay events as a result VK showing.</h3>
+                <h3>Examples with usage of overlaysContent and geometry event as a result of VK showing.</h3>
                 <aside class="example" title="Repositioning Content using CSS with overlaysContent ">
                     <pre>
                         <xmp>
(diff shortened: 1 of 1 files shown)


snianu:
There are some ReSpec errors related to `Reflect`, `ReflectOnly` & `DOMRect`. Not sure why is it not able to find DOMRect from `[geometry]`? Maybe @johanneswilm knows?

saschanaz:
> There are some ReSpec errors related to `Reflect`, `ReflectOnly` & `DOMRect`. Not sure why is it not able to find DOMRect from `[geometry]`? Maybe @johanneswilm knows?

Line 43 should be "geometry-1" instead, since that's [what specref recognizes](https://api.specref.org/bibrefs?refs=geometry-1). @sidvishnoi, do you know why that difference happens, since https://respec.org/xref shows "geometry" instead?

saschanaz:
(While you are at it, could you please remove line 46 and replace line 6 to refer to respec-w3c instead?)

snianu:
@saschanaz That fixed the `DOMRect` errors, thank you so much! I'm still getting errors for `Reflect` & `ReflectOnly`, do you know how to resolve those?

saschanaz:
Reflect/ReflectOnly have never been defined in any spec, see https://github.com/whatwg/html/issues/3238. I recommend to follow what https://github.com/WICG/conversion-measurement-api/issues/67 says until it resolves.

>Instead, imitate https://html.spec.whatwg.org/multipage/text-level-semantics.html#the-a-element:reflect by saying things like "The IDL attribute conversiondestination, impressiondata, ... must reflect the respective content attributes of the same name."

docs/virtualkeyboard/index.html
```diff
@@ -167,15 +166,24 @@ <h2>The <dfn>VirtualKeyboard</dfn> Interface</h2>
             </dt>
             <dd>
               <p>
-                When this attribute is set to `true`, a user agent MUST NOT resize its layout viewport or visual viewport.
+                A boolean, initially false. When this attribute is set to `true`, a user agent MUST NOT resize its layout viewport or visual viewport.
```
johanneswilm: false => `false`

johanneswilm (approved):

sidvishnoi:
> @sidvishnoi, do you know why that difference happens, since https://respec.org/xref shows "geometry" instead?

@saschanaz https://respec.org/xref (xref UI) makes use of data from webref, while respec uses specref. Filed https://github.com/w3c/respec-web-services/issues/224, though it's probably something that specref needs to update.

docs/virtualkeyboard/index.html
```diff
@@ -119,7 +118,7 @@ <h2>The <dfn>VirtualKeyboard</dfn> Interface</h2>
                   The method must follow these steps:
                 </p>
                 <ol>
-                    <li>If the browsing context's <a href="https://html.spec.whatwg.org/multipage/browsers.html#active-window">active window</a> does not have <a href="https://html.spec.whatwg.org/multipage/interaction.html#sticky-activation">sticky user activation</a>, abort these steps.
+                    <li>If the <a href="https://heycam.github.io/webidl/#this">this's </a><a href="https://html.spec.whatwg.org/#concept-relevant-global">relevant global object</a> is a <a href="https://html.spec.whatwg.org/multipage/window-object.html#window">Window object</a> and does not have <a href="https://html.spec.whatwg.org/multipage/interaction.html#sticky-activation">sticky user activation</a>, abort these steps.
```
yoavweiss: The links could use [cross-references](https://respec.org/docs/#referencing-terms-from-other-specifications)

yoavweiss: "the" is not needed before `this`. The "'s" following `this` would be better outside the link.

yoavweiss: Could this be called from a worker? If not, it's better to put this's relevant global in a variable, assert that it's a Window, and then abort if there's no user activation.

docs/virtualkeyboard/index.html
```diff
@@ -131,7 +130,7 @@ <h2>The <dfn>VirtualKeyboard</dfn> Interface</h2>
                       Call the system API to show the VK.
                   </li>
                   <li>
-                    When the VK is shown by the system, fire {{VirtualKeyboard/ongeometrychange()}} event.
+                    When the VK is shown by the system, the OS reports the bounding rectangle of the keyboard. Calculate the {{VirtualKeyboard/boundingRect}} which is the intersection of the keyboard rectangle reported by the OS and the layout viewport. Fire {{VirtualKeyboard/ongeometrychange()}} event to JS.
```
yoavweiss: It'd be good to define what a "bounding rectangle" is. https://drafts.csswg.org/cssom-view/#dom-element-getboundingclientrect outlines an algorithm that calculates one. It'd be good to have a similar processing model here.

docs/virtualkeyboard/index.html
```diff
@@ -143,7 +142,7 @@ <h2>The <dfn>VirtualKeyboard</dfn> Interface</h2>
                   The method must follow these steps:
                 </p>
                 <ol>
-                  <li>If the browsing context is not focused, abort these steps.
+                    <li>If the <a href="https://heycam.github.io/webidl/#this">this's </a><a href="https://html.spec.whatwg.org/#concept-relevant-global">relevant global object</a> is a <a href="https://html.spec.whatwg.org/multipage/window-object.html#window">Window object</a> and does not have <a href="https://html.spec.whatwg.org/multipage/interaction.html#sticky-activation">sticky user activation</a>, abort these steps.
```
yoavweiss: Same comments as above. May be useful to pack this in a helper algorithm.

docs/virtualkeyboard/index.html
```diff
@@ -167,15 +166,24 @@ <h2>The <dfn>VirtualKeyboard</dfn> Interface</h2>
             </dt>
             <dd>
               <p>
-                When this attribute is set to `true`, a user agent MUST NOT resize its layout viewport or visual viewport.
+                A boolean, initially false. When this attribute is set to `true`, a user agent MUST NOT resize its layout viewport or visual viewport.
+              </p>
+              <p>
+                The {{VirtualKeyboard/overlaysContent}} getter steps are to return <a href="https://heycam.github.io/webidl/#this">this's </a>{{VirtualKeyboard/overlaysContent}}.
```
yoavweiss: It'd be better to set and get a separate internal boolean variable.

yoavweiss (commented):
A few comments

docs/virtualkeyboard/index.html
snianu: Done.

docs/virtualkeyboard/index.html
snianu: Can you give me an example? I'm following https://xhr.spec.whatwg.org/#dom-xmlhttprequest-timeout

docs/virtualkeyboard/index.html
snianu: Done.

docs/virtualkeyboard/index.html
snianu: Just couple of algorithm steps so didn't create a helper

snianu:
@yoavweiss Addressed all comments. PTAL.

docs/virtualkeyboard/index.html
yoavweiss: On a second look, this is fine as is. It'd have been slightly better to define the getter as a separate concept, but that's not critical.

docs/virtualkeyboard/index.html
```diff
@@ -160,19 +178,37 @@ <h2>The <dfn>VirtualKeyboard</dfn> Interface</h2>
...
+                Setter steps are as follows:
+                <ol>
+                    <li>
+                        Let `osk` be the on-screen keyboard rectangle in CSS pixels that is provided by the OS when the VK visibility changes.
+                    </li>
+                    <li>
+                        Let `lv` be the layout viewport's bounds in CSS pixels.
+                    </li>
+                    <li>
+                        Set [=this=]'s {{VirtualKeyboard/boundingRect}} to be the intersection of `lv` and `osk` in CSS pixels.
```
yoavweiss: This algorithm feels like a getter, rather than a setter. If so, it should return a `DOMRect` but it doesn't create one. I'd expect it to do so while initializing it to the right values.

docs/virtualkeyboard/index.html
```diff
@@ -160,19 +178,37 @@ <h2>The <dfn>VirtualKeyboard</dfn> Interface</h2>
...
             </dd>
             <dt>
-              <dfn>boundingRect</dfn>
+              <p>{{VirtualKeyboard/boundingRect}}</p>
             </dt>
             <dd>
             <p>
-                The attribute reports the intersection of the VK with the layout viewport in client coordinates.
+                The attribute reports the intersection of the VK with the layout viewport in client coordinates. It is set when {{VirtualKeyboard/ongeometrychange}} event is fired.
+                Setter steps are as follows:
```
yoavweiss: There's no setter for a readonly attribute.

docs/virtualkeyboard/index.html
```diff
@@ -140,17 +154,21 @@ <h2>The <dfn>VirtualKeyboard</dfn> Interface</h2>
...
+                  </li>
+                  <li>
                     Call the system API to hide the VK.
-                </li>
-                <li>
-                    When the VK is hidden by the system, {{VirtualKeyboard/ongeometrychange()}} event is fired that contains the intersection of the keyboard geometry and layout viewport in client coordinates.
-                </li>
+                  </li>
+                  <li>
+                    When the VK is hidden by the system, the OS reports the bounding rectangle of the keyboard (which has all 0 values). Calculate the {{VirtualKeyboard/boundingRect}} which is the intersection of the keyboard rectangle reported by the OS and the layout viewport. Fire {{VirtualKeyboard/ongeometrychange()}} event to JS.
```
yoavweiss: "Calculate ...." should call an algorithm, or simply assign a 0 sized `DOMRect` (since that's the intersection IIUC)

docs/virtualkeyboard/index.html
yoavweiss: "Calculate..." should call an algorithm that actually does this calculation.

yoavweiss: This step should run [in parallel](https://html.spec.whatwg.org/#in-parallel)

docs/virtualkeyboard/index.html
```diff
@@ -116,7 +126,11 @@ <h2>The <dfn>VirtualKeyboard</dfn> Interface</h2>
                   The method must follow these steps:
                 </p>
                 <ol>
-                    <li>If the browsing context's <a href="https://html.spec.whatwg.org/multipage/browsers.html#active-window">active window</a> does not have <a href="https://html.spec.whatwg.org/multipage/interaction.html#sticky-activation">sticky user activation</a>, abort these steps.
+                  <li>
+                    Let `window` be [=this=]'s <a href="https://html.spec.whatwg.org/#concept-relevant-global">relevant global object</a>. If window is not a <a href="https://html.spec.whatwg.org/multipage/window-object.html#window">Window object</a>, then abort these steps.
```
yoavweiss: the `window` variable should be referred to as "|window|".

yoavweiss (commented):
A few more comments

docs/virtualkeyboard/index.html
snianu: oops, you are right. I'll remove it

docs/virtualkeyboard/index.html
snianu: Added a `DOMRect` object to store the intersection of the rectangles.

docs/virtualkeyboard/index.html
snianu: Done

docs/virtualkeyboard/index.html
snianu: Done.

docs/virtualkeyboard/index.html
snianu: Done.

snianu:
@yoavweiss Addressed all comments. PTAL.

docs/virtualkeyboard/index.html
```diff
@@ -107,6 +106,17 @@ <h2>The <dfn>VirtualKeyboard</dfn> Interface</h2>
...
+        <p>
+            The {{VirtualKeyboard}} object has an associated:
+        </p>
+        <dfn>boundingRect</dfn>
+        <p>
+            A {{DOMRect}}, initially has zero values.
+        </p>
+        <dfn>overlaysContent</dfn>
+        <p>
+            A boolean, initially `false`. When this attribute is set to `true`, a user agent MUST NOT resize its layout viewport or <a href="https://wicg.github.io/visual-viewport/#dom-visualviewport">visual viewport</a>.
```
yoavweiss: IntersectionObserver uses the "document's viewport", pointing to https://drafts.csswg.org/css2/#viewport
That seems more precise than "layout viewport".

docs/virtualkeyboard/index.html
```diff
@@ -116,7 +126,11 @@ <h2>The <dfn>VirtualKeyboard</dfn> Interface</h2>
                   The method must follow these steps:
                 </p>
                 <ol>
-                    <li>If the browsing context's <a href="https://html.spec.whatwg.org/multipage/browsers.html#active-window">active window</a> does not have <a href="https://html.spec.whatwg.org/multipage/interaction.html#sticky-activation">sticky user activation</a>, abort these steps.
+                  <li>
+                    Let |window| be [=this=]'s <a href="https://html.spec.whatwg.org/#concept-relevant-global">relevant global object</a>. If |window| is not a <a href="https://html.spec.whatwg.org/multipage/window-object.html#window">Window object</a>, then abort these steps.
```
yoavweiss: I *think* you can replace the abort with an assertion that |window| is a [=Window=] (because this is only exposed on Window)

Nit: You can replace the explicit links (here and elsewhere) with spec references. e.g. "Let |window| be [=this=]'s [=relevant global object=]. Assert that |window| is a {{Window}} object."

docs/virtualkeyboard/index.html
```diff
@@ -160,19 +178,42 @@ <h2>The <dfn>VirtualKeyboard</dfn> Interface</h2>
...
-              <dfn>boundingRect</dfn>
+              <p>{{VirtualKeyboard/boundingRect}}</p>
             </dt>
             <dd>
             <p>
-                The attribute reports the intersection of the VK with the layout viewport in client coordinates.
+                The attribute reports the intersection of the VK with the layout viewport in client coordinates. It is set when {{VirtualKeyboard/ongeometrychange}} event is fired. To create a {{VirtualKeyboard/boundingRect}} follow these steps:
+                <ol>
+                    <li>
+                        Let `osk` be the on-screen keyboard rectangle in CSS pixels that is provided by the OS when the VK visibility changes.
```
yoavweiss: All references to variables (osk, lv, bounds, etc) should be of the form "|var|" rather than "`var`". That enables various features such as variable highlighting, and makes it clearer that they are in fact vaariables

docs/virtualkeyboard/index.html
```diff
@@ -160,19 +178,42 @@ <h2>The <dfn>VirtualKeyboard</dfn> Interface</h2>
...
               </p>
             </dd>
             <dt>
-              <dfn>boundingRect</dfn>
+              <p>{{VirtualKeyboard/boundingRect}}</p>
             </dt>
             <dd>
             <p>
-                The attribute reports the intersection of the VK with the layout viewport in client coordinates.
+                The attribute reports the intersection of the VK with the layout viewport in client coordinates. It is set when {{VirtualKeyboard/ongeometrychange}} event is fired. To create a {{VirtualKeyboard/boundingRect}} follow these steps:
```
yoavweiss: No need to mention when it is set here. It'd be sufficient to define an algorithm here (e.g. "set a virtual keyboard bounding rect"), and then call it before the event is fired.

docs/virtualkeyboard/index.html
```diff
@@ -128,7 +142,7 @@ <h2>The <dfn>VirtualKeyboard</dfn> Interface</h2>
                       Call the system API to show the VK.
                   </li>
                   <li>
-                    When the VK is shown by the system, fire {{VirtualKeyboard/ongeometrychange()}} event.
+                    When the VK is shown by the system, <a href="https://html.spec.whatwg.org/#in-parallel">in parallel</a> the OS reports the bounding rectangle of the keyboard. Update the {{VirtualKeyboard/boundingRect}} which is the intersection of the keyboard rectangle reported by the OS and the layout viewport. Fire {{VirtualKeyboard/ongeometrychange()}} event to JS.
```
yoavweiss: You can move the "in parallel" bit up, into something like:
"""
<li>[=In parallel=], follow these steps:
* Wait for the virtual keyboard to be shown by the system.
* Call "set the virtual keyboard bounding rect" with the keyboard's OS reported bounding rectangle and the [=document=]'s [=viewport=].
* [=Fire an event=] named "geometrychange" at [=this=].

docs/virtualkeyboard/index.html
```diff
@@ -140,17 +154,21 @@ <h2>The <dfn>VirtualKeyboard</dfn> Interface</h2>
                   The method must follow these steps:
                 </p>
                 <ol>
-                  <li>If the browsing context is not focused, abort these steps.
+                  <li>
+                    Let |window| be [=this=]'s <a href="https://html.spec.whatwg.org/#concept-relevant-global">relevant global object</a>. If |window| is not a <a href="https://html.spec.whatwg.org/multipage/window-object.html#window">Window object</a>, then abort these steps.
```
yoavweiss: Same comment as above on abort and links

docs/virtualkeyboard/index.html
```diff
@@ -116,7 +126,11 @@ <h2>The <dfn>VirtualKeyboard</dfn> Interface</h2>
                   The method must follow these steps:
                 </p>
                 <ol>
-                    <li>If the browsing context's <a href="https://html.spec.whatwg.org/multipage/browsers.html#active-window">active window</a> does not have <a href="https://html.spec.whatwg.org/multipage/interaction.html#sticky-activation">sticky user activation</a>, abort these steps.
+                  <li>
+                    Let |window| be [=this=]'s <a href="https://html.spec.whatwg.org/#concept-relevant-global">relevant global object</a>. If |window| is not a <a href="https://html.spec.whatwg.org/multipage/window-object.html#window">Window object</a>, then abort these steps.
+                  </li>
+                  <li>
+                    If |window| does not have <a href="https://html.spec.whatwg.org/multipage/interaction.html#sticky-activation">sticky user activation</a>, abort these steps.
```
yoavweiss: Same nit on replacing links with references

docs/virtualkeyboard/index.html
```diff
@@ -140,17 +154,21 @@ <h2>The <dfn>VirtualKeyboard</dfn> Interface</h2>
...
+                  </li>
+                  <li>
                     Call the system API to hide the VK.
-                </li>
-                <li>
-                    When the VK is hidden by the system, {{VirtualKeyboard/ongeometrychange()}} event is fired that contains the intersection of the keyboard geometry and layout viewport in client coordinates.
-                </li>
+                  </li>
+                  <li>
+                    When the VK is hidden by the system, <a href="https://html.spec.whatwg.org/#in-parallel">in parallel</a> the OS reports the bounding rectangle of the keyboard (which has all 0 values). Create a {{DOMRect}} with 0 values and return it when {{VirtualKeyboard/boundingRect}} is queried. Fire {{VirtualKeyboard/ongeometrychange()}} event to JS.
```
yoavweiss: Same comment as above on revamping to multiple steps

docs/virtualkeyboard/index.html
yoavweiss: Also, |osk| and |lv| should be input parameters, and the caller should provide them

docs/virtualkeyboard/index.html
```diff
@@ -160,19 +178,42 @@ <h2>The <dfn>VirtualKeyboard</dfn> Interface</h2>
...
+                        Let `osk` be the on-screen keyboard rectangle in CSS pixels that is provided by the OS when the VK visibility changes.
+                    </li>
+                    <li>
+                        Let `lv` be the layout viewport's bounds in CSS pixels.
+                    </li>
+                    <li>
+                        Let `bounds` be a {{DOMRect}} object.
+                    </li>
+                    <li>
+                        Set `bounds` to be the intersection of `lv` and `osk` in CSS pixels.
```
yoavweiss: It'd be better to do something like https://www.w3.org/TR/intersection-observer/#calculate-intersection-rect-algo here. You can't really reuse it as is, because |osk| is not a DOM element...

yoavweiss (commented):
Getting closer... A few more comments

docs/virtualkeyboard/index.html
snianu: Done.

docs/virtualkeyboard/index.html
snianu: Done.

docs/virtualkeyboard/index.html
snianu: Done.

docs/virtualkeyboard/index.html
snianu: Done. Couldn't use [=viewport=] because for some reason css xref's are not working?

docs/virtualkeyboard/index.html
snianu: Done.

docs/virtualkeyboard/index.html
snianu: Done.

docs/virtualkeyboard/index.html
snianu: Done.

docs/virtualkeyboard/index.html
snianu: Done.

docs/virtualkeyboard/index.html
snianu: Done.

snianu:
@yoavweiss Addressed all comments. PTAL.

docs/virtualkeyboard/index.html
yoavweiss: @marcoscaceres - can you help on the [[=viewport=]](https://drafts.csswg.org/css2/#viewport) front?

docs/virtualkeyboard/index.html
```diff
@@ -140,47 +159,89 @@ <h2>The <dfn>VirtualKeyboard</dfn> Interface</h2>
...
             <dt>
-              <dfn>boundingRect</dfn>
+              <p>{{VirtualKeyboard/boundingRect}}</p>
             </dt>
             <dd>
             <p>
-                The attribute reports the intersection of the VK with the layout viewport in client coordinates.
+                The attribute reports the intersection of the VK with the [=document=]'s <a href="https://drafts.csswg.org/css2/#viewport">viewport</a> in client coordinates. Call {{VirtualKeyboard/set the virtual keyboard bounding rect}}.
+                <p>
+                    <p><dfn>set the virtual keyboard bounding rect</dfn></p>
```
yoavweiss, line 208: "To <dfn>set the virtual keyboard bounding rect, with |osk| (a DOMRect representing the on-screen keyboard rectangle), and |lv| (a DOMRect representing the document's viewport rectangle) as inputs, run the following steps:"

docs/virtualkeyboard/index.html
```diff
@@ -189,17 +250,17 @@ <h2>The <dfn>VirtualKeyboard</dfn> Interface</h2>
         <h2>Extensions to the <dfn>ElementContentEditable</dfn> mixin</h2>
         <pre class="idl">
             partial interface ElementContentEditable {
-                [CEReactions, Reflect, ReflectOnly=("auto","manual")] attribute DOMString virtualKeyboardPolicy;
+                [CEReactions] attribute DOMString virtualKeyboardPolicy;
```
yoavweiss, line 253: Why remove the IDL reflection rules?

yoavweiss (approved):
LGTM

I still have a few nits around turning links to refs, and one question on the IDL change, but overall this looks great. Thanks for bearing with me! :)

docs/virtualkeyboard/index.html
snianu: I was getting ReSpec errors so the recommendation (per [this ](https://github.com/w3c/editing/pull/317#issuecomment-885946703)comment) is to just mention it in the definition of the attribute and link it to the IDL. In my case I mentioned it in the `virtualKeyboardPolicy` definition:
> The virtualKeyboardPolicy is an enumerated attribute whose keywords are the string, auto, and manual. The IDL attribute must reflect the respective content attributes of the same name.

docs/virtualkeyboard/index.html
snianu: Done.

snianu:
Thank you @yoavweiss for reviewing it!

PR #320, opened 2021-08-05
add input events and clipboard api to readme

johanneswilm:
These specs are actively developed and they are listed in our charter as being on our todo-list. Therefore, they should be listed in the readme as well.


travisleithead (approved):
Looks great! Thanks for adding them!

PR #354, opened 2021-10-12
Fix variable name typo

ThomasOrlita:


travisleithead:
I don't see where the typo is... are you just wanting to swap to camel-cased variable name conventions?

ThomasOrlita:
In the sample the variable is declared as [`clipboardItem`](https://github.com/w3c/editing/blob/gh-pages/docs/clipboard-pickling/explainer.md?plain=1#L76) but then used as [`clipboard_item`](https://github.com/w3c/editing/blob/gh-pages/docs/clipboard-pickling/explainer.md?plain=1#L85).

Just fixing it to make the sample work and also to make it consistent.

travisleithead (approved):
Thanks! Non-substantive contribution.

PR #360, opened 2021-11-04
Fix minor issues

tomayac:
Fix minor issues


snianu (approved):

snianu:
Thanks for fixing these!

PR #364, opened 2021-12-01
Add a link to the EditContext Spec

johanneswilm:
I browsed around, but didn't find a way to go from our repo to the public facing spec text. Therefore this. But maybe there is a link that I just didn't see?


snianu:
@alexkeng FYI

alexkeng:
maybe we should modify https://github.com/w3c/editing/blob/gh-pages/README.md ?
right now "EditContext API" links to the explainer, while others link to specs

johanneswilm:
@alexkeng Sure - is there then still a way to quickly go to the explainer as well?

alexkeng:
well, not from the root README.md, but the explainer can be accessed from doc/EditContext/README.md, which I think makes sense (root page has the spec link, and sub folder has both the explainer link and spec link)

johanneswilm:
@alexkeng OK, fine with me.

travisleithead (approved):
Looks good!

PR #392, opened 2022-02-16
Add explainer for custom format read/write.

### LINKED ISSUE #165: Arbitrary clipboard types
As the current async clipboard API does not have a definition, extensions to it are somewhat hard to define, but here is my basic idea for arbitrary clipboard types that I think works for all parties.

The `ClipboardItem` constructor and the `ClipboardItem` `getType()` method both take a string that represents a type. As this type is a MIME type it's formatted as "`type/subtype`". The only types supported there are builtin types, such as "`text/plain`".

To extend this to allow arbitrary types, without impacting the processing of builtin types, which can be special depending on the user agent and its requirements, this proposal suggests introducing a namespace for arbitrary types. Arbitrary types will always have their data stored and read as-is, without any manipulation.

Arbitrary types are denoted by the prefix "`web `" ("`web`" followed by U+0020 SPACE) to distinguish them from builtin types. They are stored in an implementation-defined manner on the underlying OS clipboard and must not interfere with builtin types.

Expressed in JavaScript, this is how would you read an arbitrary type:

```js
const items = await navigator.clipboard.read();
const arbitraryBlob = await items[0].getType(`web ${arbitraryType}`);
```

And this is how would write one:

```js
const items = [new ClipboardItem({ `web {$arbitraryType}`: arbitraryBlob })];
navigator.clipboard.write(items);
```

@whsieh @snianu @BoCupp-Microsoft @travisleithead @johanneswilm

snianu:
Renamed the previous explainer that had the `unsanitized` option.

Closes [#165](https://github.com/w3c/clipboard-apis/issues/165)

Implementation commitment:

 * [X] Chromium (https://bugs.chromium.org/p/chromium/issues/detail?id=)
 * [X] Gecko (https://bugzilla.mozilla.org/show_bug.cgi?id=)

### DIFF
--- a/docs/clipboard-pickling/explainer.md
+++ b/docs/clipboard-pickling/explainer.md
@@ -13,7 +13,7 @@ Pickling could be used by:
 *   Web Apps to support application-specific types, for example to copy and paste within or between Google Docs and Google Sheets, (or Office 365’s Word and Excel products) using pickled formats.
 *   Developers owning both a web and native application, like SketchUp, to communicate between their applications.
 
-The existing Async Clipboard API’s re-encoding is still encouraged for use cases requiring only generic types, and easier to use as they are standardized and widely-interoperable, but pickling allows web applications with more specific or sophisticated clipboard needs to meet those needs.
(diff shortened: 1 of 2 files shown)


BoCupp-Microsoft (approved):
Consider adding a link in the current explainer to the previous version of the proposal.

PR #416, opened 2022-12-14
Change the types used on macOS

### LINKED ISSUE #412: The Web Custom formats specification does not appropriately use the macOS API
I encountered the code implementing https://github.com/w3c/editing/blob/gh-pages/docs/clipboard-pickling/explainer.md#detailed-design-discussion inside of Chromium and was pointed here. This specification does not use the macOS clipboard API appropriately and this should be fixed before shipping.

The first issue is naming. The custom format map is specified to have the format type `com.web.custom.format.map`, and the various formats are specified to have the format names `com.web.custom.format[\d+]`. This specification is not the work of [web.com](http://web.com/), and thus using the `com.web.*` prefix is inappropriate. If this is a W3C specification, a prefix of `org.w3.*` is probably the most appropriate.

The second issue is the existence of a custom format map. [Edit: This aspect of the proposal turns out to not be viable; see my followup below.] On platforms for which there is a limit on the number of formats, this might make sense. But on the Mac, types are just reverse-domain strings, and so keeping a map strikes me as a waste. Would there be an issue in not having a map type on the Mac, and instead just doing a mapping of, say, the `text/html` type to `org.w3.web-custom-format.text-html`, `my/wackytype` to `org.w3.web-custom-format.my-wackytype`, etc? There’s no reason that the native interface must look identical on all platforms, and since we already have to accomodate for their differences, we should take the opportunity to play to their strengths.

avidrissman:
This specification is not the work of web.com, and thus using the com.web.* prefix is inappropriate. Change the clipboard types used to begin with `org.w3` as this specification is theirs.

Also add macOS specific clarification.

Closes #412 

Implementation commitment:

   * I don’t believe this spec is currently implemented in WebKit, and therefore the implementation commitment for this PR is subsumed by the implementation commitment for this spec and is inapplicable
 * [X] Chromium (https://bugs.chromium.org/p/chromium/issues/detail?id=1401040)
   * A CL, https://chromium-review.googlesource.com/c/chromium/src/+/4108170, is complete
   * I don’t believe this spec is currently implemented in Gecko, and therefore the implementation commitment for this PR is subsumed by the implementation commitment for this spec and is inapplicable

### DIFF
--- a/docs/clipboard-pickling/explainer.md
+++ b/docs/clipboard-pickling/explainer.md
@@ -128,15 +128,17 @@ Note that native apps (or web apps) are free to provide an alternative form of w
 ### OS-Interaction: Format Naming
 Native applications will only be able to interact with these formats if they explicitly add support for these web custom formats. Different platforms / OS’s often have different conventions for a clipboard format’s name, so formats will be named accordingly per OS. Payloads will be unaffected/unmodified between different OS’s.
 
-On Windows & Linux, generation of clipboard formats dynamically risks exhaustion of the atom pool. On Windows, there is room for around [16,000 registered window messages and clipboard formats](https://devblogs.microsoft.com/oldnewthing/20150319-00/?p=44433). Once those are exhausted, things will start behaving erratically because window classes use the same pool of atoms as clipboard formats. Applications will not be able to register window classes until the user logs off and back on. Linux has a limitation on the atom space as well, so an approach to overcome these limitations needs to be devised.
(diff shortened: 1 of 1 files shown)


snianu:
The changes look good to me, but before we merge this change, we want to discuss with the EditingWG members to see if they have any feedback.

avidrissman:
Certainly! Please reach out if the EditingWG folks have any questions.

snianu:
We discussed this issue and no concerns were raised by the EditingWG members. Thank you @avidrissman for working on this!

snianu (approved):

avidrissman:
Thank you, @snianu, for helping shepherd this along. https://crrev.com/c/4108170 is in the commit queue and should be landed soon.

Good luck in the larger task of getting this explainer merged upstream.

PR #420, opened 2023-03-21
Add unsanitized html read/write doc.

snianu:
Add unsanitized html read/write document to facilitate discussions among stakeholders.

Implementation commitment:

 * [X] Chromium (https://bugs.chromium.org/p/chromium/issues/detail?id=1268679)

### DIFF
--- a/docs/clipboard-unsanitized/explainer.md
+++ b/docs/clipboard-unsanitized/explainer.md
@@ -6,7 +6,7 @@
 *   snianu@microsoft.com
 
 ## Introduction
-Using DataTransfer object’s setData and async clipboard write method, we are seeing interop differences in how the HTML content is sanitized and written to the clipboard. In Chromium async clipboard write method clears the clipboard content first and then writes the payload which results in overwriting the previous HTML content that was inserted by authors using DataTransfer object’s setData API. It’d be beneficial for the web authors if async clipboard and setData APIs provide the same HTML content during copy operation so round tripping is possible without any interop differences.
+Using DataTransfer object's setData and async clipboard write method, we are seeing interop differences in how the HTML content is sanitized and written to the clipboard. It'd be beneficial for the web authors if async clipboard and setData APIs provide the same HTML content during copy operation so round tripping is possible without any interop differences.
 Also creating a fragment and inlining the styles bloats the payload and [strips out the custom styles](https://drive.google.com/file/d/1Nsyp1rUKc_NF4l0n-O05snAKabHAKeiG/view) inserted by sites like Excel online that are used to preserve excel specific semantics.
 
 ## Goals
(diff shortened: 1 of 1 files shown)


docs/clipboard-unsanitized/explainer.md
```diff
@@ -0,0 +1,257 @@
+# Unsanitized HTML for Async Clipboard API
+
+
+## Author:
+*   ansollan@microsoft.com
+*   snianu@microsoft.com
+
+## Introduction
+Using DataTransfer object’s setData and async clipboard write method, we are seeing interop differences in how the HTML content is sanitized and written to the clipboard. In Chromium async clipboard write method clears the clipboard content first and then writes the payload which results in overwriting the previous HTML content that was inserted by authors using DataTransfer object’s setData API. It’d be beneficial for the web authors if async clipboard and setData APIs provide the same HTML content during copy operation so round tripping is possible without any interop differences.
```
sanketj: Should avoid usage for first person (ie. "we") in spec documents.

sanketj: Not sure it makes sense to specifically talk about the Chromium implementation of the async clipboard write API. We can talk about how the behavior is spec'ed or how the behavior differs between browsers, or it might even be fine to exclude that part. I think the introduction can be insightful enough without the implementation details.

docs/clipboard-unsanitized/explainer.md
```diff
@@ -0,0 +1,257 @@
...
+                    "text/html": html_text
+                }),
+]);
+} catch(e) {
+}
+}
+
+```
+
+## Proposal
```
sanketj, line 144: We should change this to look more like a spec, ex. https://w3c.github.io/clipboard-apis/#async-clipboard-api. Proposal is fine for an explainer, but given that we can use this document as an effective spec to understand the Chromium behavior, we should structure it like a specification.

snianu: I didn't want to duplicate all the algorithms from the clipboard API spec here as this is a minor change in `read()` & `write()` methods, and having duplicate content would most certainly result in docs going out of sync and increase confusion among web developers and UAs.
However, since this isn't an official document for clipboard APIs and has just the implementation in Chromium browser(Firefox & Safari were opposed to this feature), I can mention the IDL changes and add links to the clipboard spec for the algorithms that we are using in this feature. Thoughts?

docs/clipboard-unsanitized/explainer.md
snianu: Ack.

snianu: I can add a note here if that would be helpful, but I'm not sure if I know enough to talk about the implementation of other browsers. I'll remove the Chromium implementation details.

sanketj (approved):

PR #422, opened 2023-03-29
Additional documents for unsanitized HTML read/write

anaskim:
Additional unsanitized html read/write documents to facilitate testing of the feature.

Implementation commitment:

 WebKit (https://bugs.webkit.org/show_bug.cgi?id=)
 Chromium (https://bugs.chromium.org/p/chromium/issues/detail?id=1268679)
 Gecko (https://bugzilla.mozilla.org/show_bug.cgi?id=)

### DIFF
--- a/docs/clipboard-unsanitized/HOWTO.md
+++ b/docs/clipboard-unsanitized/HOWTO.md
@@ -5,9 +5,11 @@ Reading and writing unsanitized HTML to and from the clipboard is currently avai
 1. Download Microsoft Edge ([Canary Channel](https://www.microsoftedgeinsider.com/en-us/download/canary)).
 2. Launch Edge with the command line flag `--enable-blink-features=ClipboardUnsanitizedContentNavigate`.
 
+A similar process can be followed for Chrome ([Canary Channel](https://www.google.com/chrome/canary/)).
+
 ## Example
 
-The write method doesn't change it's shape:
+There is no change in the API shape for write() method:
 ```javascript
 const textInput = '<style>p { color: blue; }</style><p>Hello, World!</p>';
 const blobInput = new Blob([textInput], { type: 'text/html' });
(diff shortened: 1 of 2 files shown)


docs/clipboard-unsanitized/HOWTO.md
```diff
@@ -0,0 +1,40 @@
+# How to use the Async Clipboard API to read and write unsanitized HTML
+**Last updated: March, 2023**
+
+Reading and writing unsanitized HTML to and from the clipboard is currently available in Chromium-based browsers in 113 and later behind the flag `ClipboardUnsanitizedContent`.
+1. Download Microsoft Edge ([Canary Channel](https://www.microsoftedgeinsider.com/en-us/download/canary)).
```
snianu, line 5: Should we mention that similar process can be followed to test this on Chrome Canary?

docs/clipboard-unsanitized/HOWTO.md
```diff
@@ -0,0 +1,40 @@
...
+2. Launch Edge with the command line flag `--enable-blink-features=ClipboardUnsanitizedContentNavigate`.
+
+## Example
+
+The write method doesn't change it's shape:
```
snianu: Nit: Typo: The write method doesn't change **its** shape

snianu: Also would it be better to say something like " There is no change in the API shape for write() method"?

docs/clipboard-unsanitized/HOWTO.md
```diff
@@ -0,0 +1,40 @@
...
+```html
+<style>p { color: blue; }</style><p>Hello, World!</p>
+```
+
+For reference, this would be the sanitized output:
+```html
+<p style="color: blue; font-size: medium; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; white-space: normal; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;">Hello, World!</p>
+```
+
+The read method now accepts a dictionary with the `unsanitized` keyword and the `text/html` MIME type.
```
snianu: Nit: and **only** `text/html` MIME type

docs/clipboard-unsanitized/HOWTO.md
anaskim: Done.

docs/clipboard-unsanitized/HOWTO.md
anaskim: Done, thanks for the suggestion.

docs/clipboard-unsanitized/HOWTO.md
anaskim: Done.

snianu (approved):

PR #440, opened 2023-09-14
be more precise about out of scope

siusin:
Update the out of scope section as the editContext and Clipboard APIs is now in scope.


siusin:
For the record: both reviewers were satisfied with the change.

PR #453, opened 2023-10-23
Add example for read unsanitized option.

snianu:
Closes #????

Implementation commitment:


sanketj (approved):

yisibl:
Hi @snianu 

SVG has a very similar structure to HTML, is it possible to extend the unsanitized option to support SVG? This would solve the problem of missing `<style>` when copying to SVG (see https://bugs.chromium.org/p/chromium/issues/detail?id=1226150).

snianu:
@yisibl Indeed. That is exactly the intention of having an `unsanitized` option -- to allow unsanitized read, but also provide a sanitized version by-default. In this proposal we're focusing on `text/html` for now, but if we see sufficient use cases for this, then I can add it to the unsanitized list and create a separate proposal.

PR #456, opened 2023-10-27
Remove write and add more details on read.

snianu:
Closes #????

Implementation commitment:


docs/clipboard-unsanitized/explainer.md
```diff
@@ -1,16 +1,17 @@
...
-If we use the built-in sanitizer that produces an HTML fragment, the styles that get inlined bloat the payload and [strip out the custom styles](https://drive.google.com/file/d/1Nsyp1rUKc_NF4l0n-O05snAKabHAKeiG/view) inserted by sites like Excel online that are used to preserve excel specific semantics.
+DataTransfer object's `getData` and async clipboard `read()` methods have interop differences in how the HTML content is sanitized during paste operation. `getData` method returns unsanitized HTML content, but `read()` method uses the Browser sanitizer to strip out content from the HTML markup. It'd be beneficial for the web authors if async clipboard `read()` method and `getData` methods provide similar level of fidelity of HTML content during paste operations so web apps can read the HTML content written by the native apps without any interop differences such as losing formats, meta tags etc.
```
sanketj:
```suggestion
DataTransfer object's `getData` and async clipboard `read()` methods have interop differences in how the HTML content is sanitized during a paste operation. The `getData` method returns unsanitized HTML content, but the `read()` method uses the browser's sanitizer to strip out content (ex. global styles, scripts, meta tags) from the HTML markup.
```

sanketj: Instead of starting with "it will beneficial for the web authors if...", it would be good to start with what the problem is for web developers.

docs/clipboard-unsanitized/explainer.md
```diff
@@ -1,16 +1,17 @@
...
+
+If we use the built-in sanitizer that produces an HTML fragment, the styles that get inlined and bloat the payload and [strip out the custom styles](https://drive.google.com/file/d/1Nsyp1rUKc_NF4l0n-O05snAKabHAKeiG/view) inserted by sites like Excel online that are used to preserve excel specific semantics.
 
 ## Goals
-*   Preserve fidelity of the HTML format just like the legacy DataTransfer API used to read/write HTML format.
+*   Preserve fidelity of the HTML format just like the legacy DataTransfer API.
```
sanketj: Should we list parity with the DataTransfer API as a separate goal. I think preserving copy/paste fidelity deserves to be a goal on its own.

docs/clipboard-unsanitized/explainer.md
```diff
@@ -21,40 +22,52 @@ If we use the built-in sanitizer that produces an HTML fragment, the styles that
 *   Drag-and-Drop APIs.
 
 ## Additional Background
-HTML content is essential for supporting copy/paste operation of high fidelity content from native apps to web sites and vice versa, especially in sites supporting document editing. Currently it is being supported by three APIs:
+HTML content is essential for supporting copy/paste operation of high fidelity content from native apps to web sites and vice versa, especially in sites supporting document editing. Web custom formats can be used to exchange unsanitized HTML, but there are many native apps that don't have support for web custom formats, so contents copied from these apps in the HTML format would have to go through the Browser sanitizer in `read()`. This makes the `read()` method less useful as the Browser sanitizer strips out content from the HTML markup which results in format loss, bloating of payload due to inlining of style etc. Currently sites are using the DataTransfer object's `getData` method to read unsanitized HTML content, so sites do not want to regress HTML paste operation by migrating to async clipboard `read()` method.
```
sanketj: Consider moving the line about web custom formats into a separate paragraph. It is good to explain that, but it also detracts a bit from the main issue with lack of sanitization for the native HTML format.

sanketj (changes requested):
While I like that the explainer now focuses on read, which is the primary API change, it also feels important to talk about write. Copy/paste roundtripping within the browser (or hybrid apps) is an important use case as well, and that is not possible if sanitization happens on write. How should we document this?

docs/clipboard-unsanitized/explainer.md
snianu: Ack.

docs/clipboard-unsanitized/explainer.md
snianu: Ack.

docs/clipboard-unsanitized/explainer.md
snianu: Done.

snianu:
@sanketj Addressed all comments. PTAL.
For write I can create a new explainer to describe the changes we're making in the sanitization and mention it in the read explainer as a link. wdyt?

docs/clipboard-unsanitized/explainer.md
```diff
@@ -1,16 +1,18 @@
...
-If we use the built-in sanitizer that produces an HTML fragment, the styles that get inlined bloat the payload and [strip out the custom styles](https://drive.google.com/file/d/1Nsyp1rUKc_NF4l0n-O05snAKabHAKeiG/view) inserted by sites like Excel online that are used to preserve excel specific semantics.
+DataTransfer object's `getData` and async clipboard `read()` methods have interop differences in how the HTML content is sanitized during a paste operation. The `getData` method returns unsanitized HTML content, but the `read()` method uses the browser's sanitizer to strip out content (ex. global styles, scripts, meta tags) from the HTML markup.
+
+If we use the built-in sanitizer that produces an HTML fragment, the styles that get inlined and bloat the payload and [strip out the custom styles](https://drive.google.com/file/d/1Nsyp1rUKc_NF4l0n-O05snAKabHAKeiG/view) inserted by sites like Excel online that are used to preserve excel specific semantics. It'd be beneficial for the web authors if async clipboard `read()` method and `getData` methods provide similar level of fidelity of HTML content during paste operations.
```
sanketj: When introducing the problem, it is better to talk about the general issue, not the specific customer that is impacted. We can mention Excel Online (and others) in a customers/web developers section.

docs/clipboard-unsanitized/explainer.md
```diff
@@ -1,16 +1,18 @@
...
 
 ## Goals
-*   Preserve fidelity of the HTML format just like the legacy DataTransfer API used to read/write HTML format.
+*   Preserve fidelity of the HTML format.
```
sanketj:
```suggestion
*   Preserve copy/paste fidelity when reading/writing the HTML format on the clipboard.
```

docs/clipboard-unsanitized/explainer.md
```diff
@@ -21,40 +23,54 @@ If we use the built-in sanitizer that produces an HTML fragment, the styles that
 *   Drag-and-Drop APIs.
 
 ## Additional Background
```
sanketj, line 50: This sections sounds somewhat redundant with the introduction. Can they be merged?

docs/clipboard-unsanitized/explainer.md
```diff
@@ -21,40 +23,54 @@ If we use the built-in sanitizer that produces an HTML fragment, the styles that
...
+
+Web custom formats can be used to exchange unsanitized HTML, but there are many native apps that don't have support for web custom formats, so contents copied from these apps in the HTML format would have to go through the Browser sanitizer in `read()` that would result in loss of fidelity.
+
+HTML format is being supported by three APIs:
 
-### DataTransfer object's setData
-DataTransfer object can be accessed via the copy/paste event handler. It can then be used to set the clipboard data and preventDefault the browser's default copy operation. That way authors have some control over the HTML content that they want the browser to write to the native clipboard. E.g.
+### DataTransfer object's getData
+DataTransfer object can be accessed via the paste event handler. It can then be used to get the clipboard data and preventDefault the browser's default paste operation. That way authors can read the unsanitized HTML content and process the HTML markup in their document model during paste. E.g.
```
sanketj:
```suggestion
The `DataTransfer` object can be accessed via the paste event handler and `getData` can be used to get the clipboard data in a specific format. Authors can call `preventDefault` to prevent the browser's default paste action and create their own app-specific paste implementation. The `getData` API does not perform sanitization and always returns unsanitized HTML to the caller.
```

docs/clipboard-unsanitized/explainer.md
```diff
@@ -21,40 +23,54 @@ If we use the built-in sanitizer that produces an HTML fragment, the styles that
...
+  pasteTarget.contentEditable = true;
+  document.body.appendChild(pasteTarget);
+  pasteTarget.focus();
+  const result = document.execCommand("paste");
+});
+
+```
+
+### Async HTML read APIs
+This API is called via `navigator.clipboard` object and is used to read HTML to the clipboard asynchronously without depending on clipboard event or execCommand implementation. This provides more flexibility to the web authors as it doesn't need a synchronous event to access clipboard. E.g.
```
sanketj:
```suggestion
This API is called via the `navigator.clipboard` object and is used to read HTML to the clipboard asynchronously without listening for a clipboard event or calling `execCommand`. This provides more flexibility and better performance to web authors than the other APIs.
```

docs/clipboard-unsanitized/explainer.md
```diff
@@ -21,40 +23,54 @@ If we use the built-in sanitizer that produces an HTML fragment, the styles that
...
+        const customTextBlob = await clipboardItem.getType('text/html');
+        logDiv.innerText = await customTextBlob.text();
+        console.log('Text pasted.');
+        } catch(e) {
+        console.log('Failed to read clipboard');
+        }
+};
 ```
-Using any of the above mentioned APIs, web authors should be able to round trip HTML content and also be compatible with other browsers.
+Using any of the above mentioned APIs, web authors should be able to read same fidelity of HTML content.
```
sanketj:
```suggestion
All of the above-mentioned APIs should allow web authors to read HTML content with equally high fidelity.
```

docs/clipboard-unsanitized/explainer.md
```diff
@@ -21,40 +23,54 @@ If we use the built-in sanitizer that produces an HTML fragment, the styles that
...
 
-## Copy HTML text using setData
+## Paste HTML text using getData
```
sanketj: "HTML text" is a bit confusing. Consider the following:

```suggestion
## Paste HTML content using getData
```

docs/clipboard-unsanitized/explainer.md
```diff
@@ -88,61 +104,28 @@ EndFragment:00000463
 In standard html format, Safari inserts both sanitized & unsanitized version of html content. It inserts the html content provided in the setData API into the clipboard using a custom webkit format type(`com.apple.Webkit.custom-pasteboard-data`). When `getData` is called, the HTML content in the custom webkit format type is returned (makes round tripping possible).
```
sanketj: Does Safari always return data from `com.apple.Webkit.custom-pasteboard-data` when HTML is pasted? My understanding is that they only do that for same-origin sites.

docs/clipboard-unsanitized/explainer.md
```diff
@@ -88,61 +104,28 @@ EndFragment:00000463
...
 
 ### In Chromium & FF:
-During `setData` call, the HTML string is written without sanitization i.e. we don't remove tags such as `<meta>, <script>, <style>` etc from the HTML markup provided in the `setData`.
-In Chromium, the header of the HTML is hardcoded([`ui::clipboard_util::HtmlToCFHtml`](https://source.chromium.org/chromium/chromium/src/+/main:ui/base/clipboard/clipboard_util_win.cc;drc=9cc9ba08c27cb1172fb4a876ceb432f72bebfe72;l=845)) and then written to the clipboard.
+During `getData` call, the HTML string is read without sanitization i.e. we don't remove tags such as `<meta>, <script>, <style>` etc from the HTML markup provided in the `getData`.
```
sanketj:
```suggestion
When `getData` is called, the HTML string is read without sanitization i.e. global styles, script tags, meta tags are not removed from the markup.
```

docs/clipboard-unsanitized/explainer.md
```diff
@@ -88,61 +104,28 @@ EndFragment:00000463
...
-StartHTML:0000000105
-EndHTML:0000000252
-StartFragment:0000000141
-EndFragment:0000000216
-<html>
-<body>
-<!--StartFragment--><p style="color: red; font-style: oblique;">This text was copied </p><!--EndFragment-->
-</body>
-</html>
+In Chromium, the header of the HTML is hardcoded([`ui::clipboard_util::HtmlToCFHtml`](https://source.chromium.org/chromium/chromium/src/+/main:ui/base/clipboard/clipboard_util_win.cc;drc=9cc9ba08c27cb1172fb4a876ceb432f72bebfe72;l=845)) during copy and then written to the clipboard.
```
sanketj: Is this relevant for paste?

docs/clipboard-unsanitized/explainer.md
```diff
@@ -88,61 +104,28 @@ EndFragment:00000463
...
-EndHTML:0000000770
-StartFragment:0000000206
-EndFragment:0000000734
-SourceURL:file:///C:/Users/snianu.REDMOND/Downloads/index0.html
-<html>
-<body>
-<!--StartFragment--><span style="color: rgb(0, 0, 0); font-family: &quot;Times New Roman&quot;; font-size: medium; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; white-space: normal; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial; display: inline !important; float: none;">Some text</span><!--EndFragment-->
-</body>
-</html>
+Async clipboard `read()` method uses sanitizers to strip out content such as `<meta>, <style>, <script>` tags etc  from the HTML. This creates issues as it's not interop with DataTransfer's `getData` API so web authors that use `getData` (to read HTML) and async clipboard api (to read other formats) don't get the same content compared to using just the async clipboard read APIs for both HTML and other formats.
```
sanketj: I don't understand this statement. Are we suggesting that using `getData` along with the async clipboard API is important for authors? I think the main point we want to get across is that the read method only allows reading of sanitized content today.

docs/clipboard-unsanitized/explainer.md
```diff
@@ -88,61 +104,28 @@ EndFragment:00000463
...
-<body>
-<!--StartFragment--><p style="color: red; font-style: oblique;">This text was copied </p><!--EndFragment-->
-</body>
-</html>
+In Chromium, the header of the HTML is hardcoded([`ui::clipboard_util::HtmlToCFHtml`](https://source.chromium.org/chromium/chromium/src/+/main:ui/base/clipboard/clipboard_util_win.cc;drc=9cc9ba08c27cb1172fb4a876ceb432f72bebfe72;l=845)) during copy and then written to the clipboard.
 
+## Paste HTML text using async clipboard read
 ```
-Async clipboard writer API uses sanitizers to strip out content such as `<meta>, <style>, <script>` tags etc  from the HTML. This creates issues as it's not interop with DataTransfer's `setData` API so web authors that use `setData` (to write HTML) and async clipboard api (to write other formats) don't get the same content compared to using just the async clipboard write APIs for both HTML and other formats.
+<p style="color: red; font-style: oblique;">This text was copied </p>
```
sanketj: Is this example supposed to be deleted?

docs/clipboard-unsanitized/explainer.md
```diff
@@ -88,61 +104,28 @@ EndFragment:00000463
...
-                    "text/html": html_text
-                }),
-]);
-} catch(e) {}
-}
+<span style="color: rgb(0, 0, 0); font-family: &quot;Times New Roman&quot;; font-size: medium; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; white-space: normal; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial; display: inline !important; float: none;">Some text</span>
 
 ```
 
+Here the clipboard content is sanitized and tags such as `<meta>, <script>, <style>` etc are not included while pasting contents from the clipboard.
```
sanketj: A question that might come up from reading this is whether we should also update execCommand to allow unsanitized content. Should we address this as an open question?

sanketj (changes requested):

sanketj:
> @sanketj Addressed all comments. PTAL. For write I can create a new explainer to describe the changes we're making in the sanitization and mention it in the read explainer as a link. wdyt?

Thinking about this some more, I don't think we can specify that write should provide unsanitized HTML as Safari has different behavior. Perhaps it is sufficient to mention in this explainer that the `unsanitized` option in read also allows browsers that write unsanitized HTML to roundtrip HTML content better?

docs/clipboard-unsanitized/explainer.md
snianu: Ack.

docs/clipboard-unsanitized/explainer.md
snianu: Removed redundant info and merged into the intro section. I kept this section just for all the read API examples.

docs/clipboard-unsanitized/explainer.md
snianu: Ah missed that important bit of information here :) Thanks!

docs/clipboard-unsanitized/explainer.md
snianu: Clarified it a bit. We see this header info during `getData` on Windows.

docs/clipboard-unsanitized/explainer.md
snianu: Note that the default execCommand paste mutates the DOM of the active document, so it has to use the sanitizer to create the fragment which will be processed in the context of the element that has the active selection. It also fires the paste event that some sites use to access `getData` method if they want to handle the paste themselves.

docs/clipboard-unsanitized/explainer.md
snianu: Reading this part again, I think it's a bit confusing and the sanitization part is already covered in the previous sections, so will remove this section.

docs/clipboard-unsanitized/explainer.md
snianu: Removed.

snianu:
> > @sanketj Addressed all comments. PTAL. For write I can create a new explainer to describe the changes we're making in the sanitization and mention it in the read explainer as a link. wdyt?
> 
> Thinking about this some more, I don't think we can specify that write should provide unsanitized HTML as Safari has different behavior. Perhaps it is sufficient to mention in this explainer that the `unsanitized` option in read also allows browsers that write unsanitized HTML to roundtrip HTML content better?

Added this in the intro section. Thanks!

docs/clipboard-unsanitized/explainer.md
```diff
@@ -1,16 +1,20 @@
...
-If we use the built-in sanitizer that produces an HTML fragment, the styles that get inlined bloat the payload and [strip out the custom styles](https://drive.google.com/file/d/1Nsyp1rUKc_NF4l0n-O05snAKabHAKeiG/view) inserted by sites like Excel online that are used to preserve excel specific semantics.
+HTML content is essential for supporting copy/paste operation of high fidelity content from native apps to web sites and vice versa, especially in sites supporting document editing. DataTransfer object's `getData` and async clipboard `read()` methods have interop differences in how the HTML content is sanitized during a paste operation. The `getData` method returns unsanitized HTML content, but the `read()` method uses the browser's sanitizer to strip out content (ex. global `<style>`s, `<script>`s, `<meta>` tags) from the HTML markup which results in format loss, bloating of payload due to inlining of styles etc.
```
sanketj: What does "etc." refer to here?

Can we also add an example for each problem that we are calling out? A bulleted list can be used if that provides better formatting.

docs/clipboard-unsanitized/explainer.md
```diff
@@ -1,16 +1,20 @@
...
+
+Currently sites that are using the DataTransfer object's `getData` method to read unsanitized HTML content, do not want to regress HTML paste operation by migrating to async clipboard `read()` method. It'd be beneficial for the web authors if async clipboard `read()` method and `getData` methods provide similar level of fidelity of HTML content during paste operations. This would also allow browsers that write unsanitized HTML content into the clipboard to roundtrip HTML content better.
```
sanketj:
```suggestion
These problems mean that web developers may not get the same HTML paste quality and performance with the async clipboard `read` API as they do with the DataTransfer object's `getData` method. This proposal aims to solve these problems so that the `read` can work just as well as `getData` when pasting HTML content.
```

sanketj: For:
"This would also allow browsers that write unsanitized HTML content into the clipboard to roundtrip HTML content better."
Should we list this in the goals section below instead?

docs/clipboard-unsanitized/explainer.md
```diff
@@ -1,16 +1,20 @@
...
+
+Web custom formats can be used to exchange unsanitized HTML if both source and target apps have support for it, but there are many native apps that don't have support for web custom formats, so contents copied from these apps in the HTML format would have to go through the Browser sanitizer in `read()` that would result in loss of fidelity.
```
sanketj: It might be better to move this into an alternatives considered section. Thoughts?

docs/clipboard-unsanitized/explainer.md
```diff
@@ -85,64 +100,14 @@ EndFragment:00000463
...
 
 ## Proposal
 
-With this new proposal, we will be introducing a new `unsanitized` parameter in the [read()](https://w3c.github.io/clipboard-apis/#dom-clipboard-read) method so the content is round trippable i.e. `read()` would return the content without any sanitization. On [write](https://w3c.github.io/clipboard-apis/#dom-clipboard-write) method call, we will always write a well-formed HTML document if `text/html` is provided in the [ClipboardItem](https://w3c.github.io/clipboard-apis/#clipboard-item-interface).
+With this new proposal, we will be introducing a new `unsanitized` parameter in the [read()](https://w3c.github.io/clipboard-apis/#dom-clipboard-read) method so the HTML content can be read without any loss of information i.e. `read()` would return the content without any sanitization.
```
sanketj: Should we update `read()` to pass the unsanitized parameter?

docs/clipboard-unsanitized/explainer.md
```diff
@@ -214,32 +151,17 @@ const blobOutput = await clipboardItems[0].getType('text/html');
...
+Websites or native apps are already reading unsanitized content via DataTransfer APIs using `getData()` method. In this proposal, web authors are required to explicitly specify `unsanitized` option in the async clipboard `read()` method to access the raw `text/html` content from the clipboard. This feature uses async clipboard API that has a [user gesture requirement](https://w3c.github.io/clipboard-apis/#check-clipboard-read-permission) on top of [existing](https://github.com/dway123/clipboard-pickling/blob/main/explainer.md#permissions) async clipboard API security measures to mitigate security and privacy concerns.
 
 For more details see the [security-privacy](https://github.com/MicrosoftEdge/MSEdgeExplainers/blob/main/ClipboardAPI/tag-security-privacy-clipboard-unsanitized-read.md) doc.
 
+[Here](https://docs.google.com/document/d/1QLt50Q8UnlQksVltZ2PNkDZVdk9N58Pq7P0lzGTKh-U/edit?usp=sharing) is a threat model document for this feature.
```
sanketj: This document was made for Chromium T&S reviews. I don't think we should include a link in the explainer without sanitization (pun intended :) ). I don't think it is strictly necessary here given that you have described the security considerations elsewhere. So I would actually suggest just leaving it out.

sanketj (changes requested):

docs/clipboard-unsanitized/explainer.md
snianu: Remove etc and added examples for all the problems mentioned.

snianu:
@sanketj Addressed all comments. PTAL, thanks!

docs/clipboard-unsanitized/explainer.md
```diff
@@ -1,16 +1,43 @@
...
+After `read()` was called with the default sanitizer, the HTML markup returned was:
+
+```js
+<p style="background-color: blue; color: rgb(0, 0, 0); font-size: medium; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; white-space: normal; text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;">html text</p>
+
+```
+
+In the above example, `script`, `style` tags were removed and style associated with `<p>` element was inlined.
+
+These problems mean that web developers may not get the same HTML paste quality and performance with the async clipboard `read` API as they do with the DataTransfer object's `getData` method. This proposal aims to solve these problems so that the `read` can work just as well as `getData` when pasting HTML content.
```
sanketj, line 35: It's not clear how stripping scripts and inlining styles causes fidelity loss. Is it because meta tags get stripped, or because not all styles are inlined?

docs/clipboard-unsanitized/explainer.md
```diff
@@ -248,6 +191,9 @@ This requirement is now enforced for the Async Clipboard API overall. It may be
 ### Permissions
 Due to concerns regarding permission fatigue and comprehensibility, and due to the limited utility of a permission, no new permission would be implemented for unsanitized clipboard. Given that Clipboard API read and write are already permitted, unsanitized clipboard read and write will be permitted as-is.
 
+## Alternatives considered 
+Web custom formats can be used to exchange unsanitized HTML if both source and target apps have support for it, but there are many native apps that don't have support for web custom formats, so contents copied from these apps in the HTML format would have to go through the Browser sanitizer in `read()` that would result in loss of fidelity.
```
sanketj:
```suggestion
Web custom formats can be used to exchange unsanitized HTML if both source and target apps have support for it, but there are many native apps that don't have support for web custom formats, so contents copied from these apps in the HTML format would have to go through the browser's sanitizer in `read()` that would result in loss of fidelity.
```

sanketj (changes requested):

docs/clipboard-unsanitized/explainer.md
snianu: I think it's the custom styles that get stripped out which results in fidelity loss. Meta tags do not affect the fidelity of HTML content.

docs/clipboard-unsanitized/explainer.md
snianu: Done.

docs/clipboard-unsanitized/explainer.md
snianu: Done.

sanketj (approved):

PR #457, opened 2023-12-18
Don't enable execCommand commands in EditContext

### LINKED ISSUE #71: `Document.execCommand`, `Document.queryCommandValue`, `Document.queryCommandState`, `Document.queryCommandEnabled`, `Document.queryCommandSupported`, `Document.queryCommandIndeterm` behavior when an EditContext editing host has focus
I think `Document.execCommand` should return `false` and does nothing. `Document.queryCommandEnabled` and `Document.queryCommandSupported` should return `false`. `Document.queryCommandState` should return `false`. `Document.queryCommandValue` should return empty string. And `Document.queryCommandIndeterm` should return `false` even though may be opposite from the name, but I think browsers returns `false` if the command is not available.

dandclark:
Closes https://github.com/w3c/edit-context/issues/71.

Update execCommand spec per resolution of https://github.com/w3c/edit-context/issues/71#issuecomment-1806317341 to indicate that execCommand commands are not enabled in EditContext, except for commands that are enabled regardless of whether the command is performed in an editable region.

 * [X] Confirmed there are no ReSpec/BikeShed errors or warnings.
 * [X] Modified Web platform tests: https://github.com/web-platform-tests/wpt/commit/5411d9cadd5b8dcdd7c84c053f8420b105fc2a90

Implementation commitment:

 * [ ] WebKit (haven't yet implemented EditContext)
 * [X] Chromium: https://chromium-review.googlesource.com/c/chromium/src/+/4975435
 * [ ] Gecko (haven't yet implemented EditContext)

### DIFF
--- a/docs/execCommand/index.html
+++ b/docs/execCommand/index.html
@@ -590,9 +590,15 @@
           "command">commands</a> defined here are <a href=
           "#enabled">enabled</a> if the <a href="#active-range">active
           range</a> is not null, its [=range/start node=] is either <a href=
-          "#editable">editable</a> or an [=editing host=], its [=range/end
-          node=] is either <a href="#editable">editable</a> or an [=editing
-          host=], and there is some [=editing host=] that is an
+          "#editable">editable</a> or an [=editing host=], the <a href="#editing-host-of">editing
+          host of</a> its [=range/start node=] is not an
+          <a href="https://w3c.github.io/edit-context/#dfn-editcontext-editing-host">EditContext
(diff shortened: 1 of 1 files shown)


snianu (approved):
Could we add some WPT tests for this?

snianu:
> Could we add some WPT tests for this?

nvm. I just realized that you added tests in third_party/blink/web_tests/external/wpt/editing/edit-context/edit-context-execCommand.tentative.https.html

PR #515, opened 2025-12-08
add a draft charter for the 2025 recharter

siusin:
Please review:

Name of the new spec: Caret Movement
Abstract: This specification aims to improve the user experience of Bidirectional text selection in scenarios where LTR and RTL text are both used, as well as in other complex situations such as arrow navigation of the cursor around inline images, and text input within contenteditable elements.

/cc @ragoulik

### DIFF
--- a/charter-drafts/editing-2025.html
+++ b/charter-drafts/editing-2025.html
@@ -1,0 +1,572 @@
+<!DOCTYPE html>
+<html lang="en-US">
+  <head>
+    <meta charset="utf-8">
+
+    <title>[draft] Web Editing Working Group Charter</title>
+
+    <link rel="stylesheet" href="https://www.w3.org/2005/10/w3cdoc.css" type="text/css" media="screen">
+    <link rel="stylesheet" type="text/css" href="https://www.w3.org/OldGuide/pubrules-style.css">
+    <link rel="stylesheet" type="text/css" href="https://www.w3.org/2006/02/charter-style.css">
+    <style>
+      main {
+        max-width: 60em;
+        margin: 0 auto;
+      }
+
+      ul#navbar {
+        font-size: small;
+      }
+
+      dt.spec {
+        font-weight: bold;
+      }
+
+      dt.spec new {
+        background: yellow;
+      }
+
+      ul.out-of-scope > li {
+        font-weight: bold;
+      }
+
+      ul.out-of-scope > li > ul > li{
+        font-weight: normal;
+      }
+
+      .issue {
+        background: cornsilk;
+        font-style: italic;
+      }
+
+      .todo {
+        color: #900;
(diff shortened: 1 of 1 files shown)


charter-drafts/editing-2025.html
```diff
@@ -0,0 +1,572 @@
...
+        <p>
+          This group will seek to make decisions through consensus and due process, per the <a href="https://www.w3.org/Consortium/Process/#Consensus">W3C Process Document (section 5.2.1, Consensus)</a>. Typically, an editor or other participant makes an initial proposal, which is then refined in discussion with members of the group and other reviewers, and consensus emerges with little formal voting being required.</p>
+        <p>
+           However, if a decision is necessary for timely progress and consensus is not achieved after careful consideration of the range of views presented, the Chairs may call for a group vote and record a decision along with any objections.
+        </p>
+        <p>
+          To afford asynchronous decisions and organizational deliberation, any resolution (including publication decisions) taken in a face-to-face meeting or teleconference will be considered provisional.
+
+          A call for consensus (CfC) will be issued for all resolutions (for example, via email, GitHub issue or web-based survey), with a response period from <span id='cfc'>one week</span>, depending on the chair's evaluation of the group consensus on the issue.
```
johanneswilm, line 415: @siusin Is it possible to adjust the wording here that a CfC is issued implicitly automatically with every resolution to make it clear that we do not start a separate manual process of issuing a CfC with every resolution?

johanneswilm (approved):

johanneswilm:
On the new spec:

TPAC 2025:
> support for Bidirectional(presence of both ltr and rtl) text: Selection and cursor navigation – where should we start?
Rakesh
We dont have github issue or proposal for bidirectional support. Still in ideation phase
Concept of proposal - today a common practice for people to use LTR and RTL text in same sentence in same world. 
In chromium based browsers, in selection part, it has both combination. We want to improve UX to support such combinations
We plan to have  a proposal, don’t have much as of now. Just wanted to pitch the idea
Johannes
We already spoke about making new spec, not much about directions
Caret movement in general is oe of things, close to selection but not the same
There are other questions - editor inside editor does it go around it ? should go in editing spec
Given another issue about caret movement all should go into new spec - with a specific section
We solved it 10 years ago - caret wouldn’t go certain places. It was less of a need. But a spec would be good with new browsers coming along. 
Are you proposing a new spec, can we have another section with carent move around svgs, images and other problematic elements
Rakesh
Would be good idea
Johannes
Need a name for spec. - Caret movement spec
Smaug
Sounds reasonable
Rakesh:
Sounds good
Johannes
Resolution - have new spec “Caret movement”. Rakesh will be the editor. I can share more issues around the caret movement. Looking forward for some proposals

Issue #516, opened 2025-12-10
Safari Desktop: Adding an empty span to an otherwise empty editable element, makes the element unselectable


michael:
CONTEXT: As an editor developer I need a way to position popovers relative to a collapsed cursor. That's why I'm rendering an additional span at the cursor position with an anchor-name, so I can use CSS anchor positioning to position a popover/prompt (e.g. for adding a new link).

BUG: With the extra span in place, in Safari the field(s) are no longer selectable.
In Chrome and Firefox it works just fine.

See Repro: https://w3c.github.io/editing/repros/516.html

CONTEXT:

Behavior of empty fields broken on Safari at https://svedit.dev (v0.5.0)

https://github.com/user-attachments/assets/9163e5fe-be8f-4ba5-803f-c282b66c72de

Popovers anchored at selection highlight span, using CSS Anchor positioing:

https://github.com/user-attachments/assets/e07dd78d-50af-4735-a8d4-b731721be914

michael:
Likely related: #499

johanneswilm:
Call 2025-12-11:
> 08:24 <whsieh> johanneswilm: Michael will upload a video. we'll keep the Agenda+ and see if we have an idea for a cleaner solution 
08:24 <whsieh> Michael: ideally — would want to anchor to the selection
08:24 <whsieh> (similar to how you want to anchor to a DOM node)
08:25 <whsieh> johanneswilm: might take a while to develop exactly what CSS anchor positioning relative to selection would look like. seems interesting to pursue
08:25 <whsieh> johanneswilm: let's see if opacity: 0 is a viable workaround

michael:
> 08:24 johanneswilm: Michael will upload a video. we'll keep the Agenda+ and see if we have an idea for a cleaner solution

Done, video is part of the issue description now.

> 08:25 johanneswilm: let's see if opacity: 0 is a viable workaround

`opacity: 0` to hide the br works but is not an option, as this will still render the cursor in the wrong place when placeholders are involved. See https://github.com/w3c/editing/issues/498.

> 08:24 Michael: ideally — would want to anchor to the selection

Created a separate issue for discussing that feature: https://github.com/w3c/editing/issues/518

michael:
Official bug report filed: https://bugs.webkit.org/show_bug.cgi?id=304143

johanneswilm:
Call 2026-02-12:

> 08:11 <whsieh> WK bugs have been filed
[...]

PR #517, opened 2025-12-10
Rename 512.html to 513.html and add repro for #516

michael:


michael:
@johanneswilm, pls review and merge.

Issue #531, opened 2026-04-08
Safari: `height: 100%` changes layout when the parent height is defined via `aspect-ratio`


michael:
Not related to editing, but tracking it here together with a repro.

Consider this markup:

```html
<div style="width: 500px; aspect-ratio: 7882 / 5257;">
  <div style="height: 100%"></div>
</div>
```

In Safari (both on desktop and mobile), the inner div with `height: 100%` leads to a few extra pixels added to the height of the inner and outer div. I suspect some wrong height calculation.

See Repro: https://w3c.github.io/editing/repros/531.html (pending PR being merged)

michael:
Filed this at the Webkit bugtracker directly. Closing here as it's off-topic.

New ticket: https://bugs.webkit.org/show_bug.cgi?id=311833

PR #532, opened 2026-04-08
Repro for #531

michael:


michael:
Closing in favor of:

https://bugs.webkit.org/show_bug.cgi?id=311833

and 

https://cdn.letsken.com/repros/001-safari-height-100-percent.html

PR #534, opened 2026-05-01
restore mitigations and residual risk text. editing-2025.html

tantek:
restore text on mitigations and clear description of the residual risk from previous/current charter

This is a charter edit. Removed the repo PR edit boilerplate which does not make sense for a charter edit.


siusin (approved):
Thank you!

